PR #1, opened 2023-05-03
JWS Update

nodh:
Update JWS data classes to provide more flexibility, e.g. for implementing protocols for issuance.


JesusMcCloud (approved):
👍

PR #4, opened 2023-05-23
Code Organization

nodh:
Bump to 2.0.0 due to extensive code changes:

- Split out implementation of protocols into `vclib-aries` and `vclib-openid` artifacts
- Implement [OpenID 4 Verifiable Credential Issuance](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html)
- Remove notion of "attribute names" and use "attribute types" only


vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/IssuerService.kt
```diff
@@ -0,0 +1,96 @@
...
+        )
+        return IssuerMetadata(
+            issuer = publicContext,
+            credentialIssuer = publicContext,
+            authorizationServer = "https://eid.egiz.gv.at/",
+            authorizationEndpointUrl = "$publicContext/authorize",
+            tokenEndpointUrl = "$publicContext/token",
+            credentialEndpointUrl = "$publicContext/credential",
+            supportedCredentialFormat = arrayOf(credentialFormat),
+            displayProperties = arrayOf(DisplayProperties(name = "ID Austria Credential", locale = "de"))
```
JesusMcCloud, line 39: overly specific hardcoded values?

JesusMcCloud:
I'll try to figure out how common `implementation` dependencies can be declared only once, but still shared.
Looks very cleanly separated otherwise!

JesusMcCloud:
Erros made me realise, that there are case Diffecences between the modules ("KMM", vs. "Kmm")

JesusMcCloud:
> Errors made me realise, that there are case differences between the modules ("KMM", vs. "Kmm")

still open. to be more precise and less confusing, we still have:
 * VcLibAries`Kmm`
 * VcLibOpenId`Kmm`
 * VcLib`KMM`

Is that intentional?

nodh:
In my opinion, `Kmm` looks cleaner, so we should change it.

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OpenIdConstants.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    const val ID_TOKEN = "idToken"
+
+    const val GRANT_TYPE_CODE = "code"
+
+    const val TOKEN_PREFIX_BEARER = "Bearer "
+
+    const val TOKEN_TYPE_BEARER = "bearer"
+
+    const val PATH_WELL_KNOWN_CREDENTIAL_ISSUER = "/.well-known/openid-credential-issuer"
```
JesusMcCloud, line 13: :+1:

PR #6, opened 2023-06-21
Update OIDC SIOPv2 Implementation

nodh:
Refactors the classes for verifier and holder/wallet, implements more response types and variants.


vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/IdTokenTypeSerializer.kt
JesusMcCloud, line 1: :+1:

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/IssuerMetadata.kt
```diff
@@ -21,7 +23,7 @@ data class IssuerMetadata(
      * OID4VCI: REQUIRED. The Credential Issuer's identifier.
      */
     @SerialName("credential_issuer")
-    val credentialIssuer: String,
+    val credentialIssuer: String? = null,
```
JesusMcCloud, line 26: Not trying to over-engineer stuff here, but would it not be nice to subclass for better nullability? (also applies to other parts of the codebase)

nodh: As discussed offline, this is probably not that easy due to the complexity of possible features/extensions for OAuth

JesusMcCloud (approved):

PR #7, opened 2023-06-29
add source links to docs

JesusMcCloud:
please verify locally that it works by invoking `./gradlew clean dokkaHtmlMultiModule` and then navigating around a bit:

Html Doc:
image

Result after clicking on "(source)":
image


nodh (approved):
lgtm :-)

PR #8, opened 2023-06-30
use new KmmResult

JesusMcCloud:
(needs some time  until KmmResult actually lands in maven central)

### DIFF
--- a/vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
+++ b/vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
@@ -114,7 +114,7 @@ class OidcSiopWallet(
         val params = kotlin.runCatching {
             Url(it).parameters.flattenEntries().toMap().decodeFromUrlQuery<AuthenticationRequestParameters>()
         }.getOrNull()
-            ?: return KmmResult.failure(OAuth2Exception(Errors.INVALID_REQUEST))
+            ?: return KmmResult.failure<AuthenticationResponseResult>(OAuth2Exception(Errors.INVALID_REQUEST))
                 .also { Napier.w("Could not parse authentication request") }
         return extractRequestObject(params)
             ?.let { createAuthnResponse(it) }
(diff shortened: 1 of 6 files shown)


vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
```diff
@@ -113,7 +113,7 @@ class OidcSiopWallet(
         val params = kotlin.runCatching {
             Url(it).parameters.flattenEntries().toMap().decodeFromUrlQuery<AuthenticationRequestParameters>()
         }.getOrNull()
-            ?: return KmmResult.failure(OAuth2Exception(Errors.INVALID_REQUEST))
+            ?: return KmmResult.failure<AuthenticationResponseResult>(OAuth2Exception(Errors.INVALID_REQUEST))
```
nodh, line 116: Why do we need the type parameter here, and not in other functions?

nodh (changes requested):
Waiting for the CI job to run, and one comment

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
JesusMcCloud: The new KmmResult requires type declaration also for the failure case since the <Nothing>-bounded subclass is gone. Automatic inference fails here. In this particular context I'd need to really think through why the declared return type of the function is not enough

JesusMcCloud:
Blocked by Kotlin bug [KT-61382](https://youtrack.jetbrains.com/issue/KT-61382)

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
JesusMcCloud: This will probably not be necessary anymore with K2 (i.e. kotlin 2.0).
The frontend simply fails to infer the type and there is nothing we can do. resolved?

JesusMcCloud:
the [KT-61382](https://youtrack.jetbrains.com/issue/KT-61382) workaround makes it work, however, merging this one results in an issue trying to register `TestCredential` in the oidc testcase for serialization, which I do not understand

PR #9, opened 2023-08-03
Add certificates in ISO COSE structures

nodh:
Opening a PR to not intervene with build black magic


vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Validator.kt
```diff
@@ -295,14 +299,17 @@ class Validator(
      *
      * @param it The [IssuerSigned] structure from ISO 18013-5
      */
-    fun verifyIsoCred(it: IssuerSigned, issuerKey: CoseKey): Verifier.VerifyCredentialResult {
+    fun verifyIsoCred(it: IssuerSigned, issuerKey: CoseKey?): Verifier.VerifyCredentialResult {
```
JesusMcCloud, line 302: function called `verify` turns into NOOP with null key?

nodh: it returns `InvalidStructure`, three lines below

JesusMcCloud (approved):
I changed the target branch to`feature/mdl-cbor` and updated accordingly

PR #12, opened 2023-08-24
make cert optional in CryptoPublicKey

JesusMcCloud:
please test on the target branch, as I did not want to replicate the fixes to the test cases, since they were not cherry-pickable. hence I am directly assigning this to you, @nodh


nodh:
But which fixes are needed?

PR #25, opened 2023-10-24
Credential Representations

nodh:
This series of commits lifts the 1-1 relation between credential and its format. Now, any credential can be represented in any format (that is SD-JWT, Plain JWT, ISO MDOC), when supported by the implementation of the `IssuerCredentialDataProvider`.
As a consequence, the Wallet may request the same credential type from the backend in all three representations and store them. When a Verifier later on requests the credential type in a specific representation, the Wallet can create a presentation in the requested format.
In other news, the implementation of a custom `CredentialScheme` got a bit easier, as the `getClaims()` method from `CredenitalSubject` has been dropped (introduced in 3.1.0).
Please note that this applies to the OpenId protocol implementations only, as upgrading the ARIES protocols seems not worth the effort.


vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopVerifier.kt
```diff
@@ -170,42 +172,26 @@ class OidcSiopVerifier(
...
-            else -> ConstraintField(
-                path = arrayOf("$.type"),
-                filter = ConstraintFilter(
-                    type = "string",
-                    pattern = "AnyCredential",
-                )
-            )
+        val typeConstraint = credentialScheme?.let {
+            when (credentialRepresentation) {
+                ConstantIndex.CredentialRepresentation.PLAIN_JWT -> it.vcConstraint()
```
JesusMcCloud, line 188: We have both functions for all credentials, so we can encode mDL as W3C VCs for debugging, right?!

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/data/AriesGoalCodeParser.kt
JesusMcCloud, line 1: Sounds like a villain of the power rangers

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/iso/MobileDrivingLicenceJwsNamespace.kt
```diff
@@ -11,7 +11,7 @@ import kotlinx.serialization.encodeToString
  */
 @Serializable
 data class MobileDrivingLicenceJwsNamespace(
-    @SerialName(IsoDataModelConstants.NAMESPACE_MDL)
+    @SerialName("org.iso.18013.5.1")
```
JesusMcCloud, line 14: Why?

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopVerifier.kt
nodh: Yes, we can

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/iso/MobileDrivingLicenceJwsNamespace.kt
nodh: Because it's the only usage of this string value left now

JesusMcCloud (approved):

PR #33, opened 2024-01-30
Update errorhandling for Wallets, SIOP QR Codes

nodh:
Implements improvements for Wallet Apps:

- Passes exceptions from native crypto service implementations up to callers of the `JwsService`
- Adds QR Code use case for SIOPv2
- Store entries are serializable

In my opinion we should release version 3.4.0 afterwards.


vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopVerifier.kt
```diff
@@ -95,6 +95,48 @@ class OidcSiopVerifier(
         )
     }
 
+    private val containerJwt = FormatContainerJwt(algorithms = arrayOf(JwsAlgorithm.ES256.identifier))
```
JesusMcCloud, line 98: Why is the algo hardcoded here?

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
```diff
@@ -231,9 +235,10 @@ class OidcSiopWallet(
         )
         val jwsPayload = idToken.serialize().encodeToByteArray()
         val jwsHeader = JwsHeader(algorithm = JwsAlgorithm.ES256)
```
JesusMcCloud, line 237: should we not take this opportunity to get rid of all the hardcoded stuff and at least make it a constructor property, that defaults to ES256? I'm not saying this PR should do it, but before we release 3.4.0, those hardcoded values should go.

JesusMcCloud (commented):
Looks good except for the remnants of hardcoded P-256-based algos. But that's a separate issue and should be done in a separate PR (but still before 3.4.0, becuase now we have a CryptoService that supports all curves, but other pieces of the puzzle are hardcoded to only fit P-256.

JesusMcCloud (approved):
Looks good except for the remnants of hardcoded P-256-based algos. But that's a separate issue and should be done in a separate PR (but still before 3.4.0, becuase now we have a CryptoService that supports all curves, but other pieces of the puzzle are hardcoded to only fit P-256.

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
nodh: But in `JwsHeader`, the `algorithm` parameter is required, i.e. non-null.

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopVerifier.kt
nodh: Fixed that in recent commits, for several usages of hardcoded algorithms in the OID4VCI/SIOP implementations.

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/jws/JwsService.kt
```diff
@@ -107,36 +126,33 @@ class DefaultJwsService(private val cryptoService: CryptoService) : JwsService {
...
+        val enc = header.encryption
+            ?: return KmmResult.failure(IllegalArgumentException("No encryption in JWE header"))
+        val epk = header.ephemeralKeyPair
+            ?: return KmmResult.failure(IllegalArgumentException("No epk in JWE header"))
+        val z = cryptoService.performKeyAgreement(epk, alg).getOrElse {
+            Napier.w("No Z value from native code", it)
+            return KmmResult.failure(it)
+        }
+        val kdfInput = prependWithAdditionalInfo(z, enc, header.agreementPartyUInfo, header.agreementPartyVInfo)
         val key = cryptoService.messageDigest(kdfInput, Digest.SHA256).getOrElse {
```
JesusMcCloud, line 143: 🤔

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/jws/JwsService.kt
```diff
@@ -165,23 +180,22 @@ class DefaultJwsService(private val cryptoService: CryptoService) : JwsService {
         val z = cryptoService.performKeyAgreement(ephemeralKeyPair, recipientKey, jweAlgorithm)
             .getOrElse {
                 Napier.w("No Z value from native code", it)
-                return null
+                return KmmResult.failure(it)
             }
         val kdf = prependWithAdditionalInfo(z, jweEncryption, null, null)
         val key = cryptoService.messageDigest(kdf, Digest.SHA256).getOrElse {
```
JesusMcCloud, line 186: 🤔

JesusMcCloud (commented):
Are the hardcoded SHA-256 digests a problem?

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/jws/JwsService.kt
nodh: This is fixed per spec https://datatracker.ietf.org/doc/html/rfc7518#section-4.6.2

JesusMcCloud (approved):

PR #34, opened 2024-02-01
Moved credentialRepresentation from parameter to function

n0900:
In order to not having to instantitate an `OidcSiopVerifier` for every `credentialRepresentation` we can move it from a class parameter to a function parameter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 # Changelog
 
 Release 3.4.0:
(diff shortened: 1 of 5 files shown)


nodh:
I did extend your code a bit, what do you think?

n0900:
> I did extend your code a bit, what do you think, @n0900 ?

You are right. Refactoring `metadata` and `FormatHolder` looks way better.

nodh (approved):

PR #35, opened 2024-02-05
remove `iosDisabled` switch

JesusMcCloud:


nodh (approved):
lgtm

stefanmore:
for the record, since Release 5.9.0 (#414) the switch `disableAppleTargets` exists instead:

> Make it possible to disable all apple targets by setting Gradle property `disableAppleTargets=true` (either through `gradle.properties`/`local.properties` or as env variable)

PR #36, opened 2024-02-15
add mirroring action

JesusMcCloud:
This PR adds the mirroring action to this repo.
Please do check if more files should be pruned before we merge this into main. E. g. do we really want to build it and consume minutes on a mac runner. isn't testing enough?

Also: check the repository settings for the variables used in the action

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -1,0 +1,61 @@
+# This is a basic workflow to help you get started with Actions
+
+name: Mirror
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the "main" branch
+  push:
+    branches: [ "main" ]
+
+  # Allows you to run this workflow manually from the Actions tab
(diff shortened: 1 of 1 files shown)


.github/workflows/mirror.yml
```diff
@@ -0,0 +1,61 @@
+# This is a basic workflow to help you get started with Actions
```
nodh, line 1: shouldn't we remove this comments?

nodh (changes requested):

nodh:
> This PR adds the mirroring action to this repo. Please do check if more files should be pruned before we merge this into main. E. g. do we really want to build it and consume minutes on a mac runner. isn't testing enough?

Probably we should remove every action, since they're running here anyways.

.github/workflows/mirror.yml
JesusMcCloud: yes, i should

JesusMcCloud: removed

PR #37, opened 2024-03-02
EC Point Compression and Correct multibase encoding

JesusMcCloud:
All thanks to updated KMP Crypto.

Also fixes an issue (which until now never struck) caused by the vclib-conventions Plugin file (not class name) being called `Plugin.kt` just as the gradle-convention-plugin's Plugin file (which has a different class name).


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+Release NEXT:
+- Update to latest KMP Crypto
+  - Introduces correct mulitbase encoding
+  - EC Point Compression
+  - **THIS IS A BREAKING CHANGE WRT. SOME ASPECTS OF SERIALIZATION**
```
nodh, line 7: We should at least document this breaking change here (essentially plain byte encoding vs var int encoding, right?)

nodh (changes requested):

CHANGELOG.md
JesusMcCloud: you mean more elaboration?

JesusMcCloud: better now?

nodh (approved):

PR #45, opened 2024-03-22
Thorough request object retrieval algorithm

acrusage:
Improved algorithm for retrieving AuthenticationResponseParameters as asked for in https://github.com/a-sit-plus/compose-wallet-app/pull/61#discussion_r1534006344

### DIFF
--- a/vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
+++ b/vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
@@ -172,10 +172,6 @@ class OidcSiopWallet(
                 .decodeFromUrlQuery<AuthenticationRequestParameters>()
         }.getOrNull() ?: kotlin.runCatching {
             // maybe it's a url that yields the request object in some other way
-            // currently supported in order of priority:
-            // 1. use redirect location as new starting point if available
-            // 2. use resonse body as new starting point
-            // - maybe it's just a jws that needs to be parsed, but let's also support a url there
             val url = Url(input)
             val candidates = requestObjectCandidateRetriever.invoke(url)
             var result: AuthenticationRequestParameters? = null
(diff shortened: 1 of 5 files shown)


JesusMcCloud (changes requested):
The implementation looks really good, but it would be nice to have both cases tested and not only the baseline one we hade before (even though it is important to keep it as regression test)

acrusage:
> The implementation looks really good, but it would be nice to have both cases tested and not only the baseline one we hade before (even though it is important to keep it as regression test)

I might need help with that, I'm not sure how to provide a dummy response to a HttpClient.get request

acrusage:
> The implementation looks really good, but it would be nice to have both cases tested and not only the baseline one we hade before (even though it is important to keep it as regression test)

fixed in commit 19b461408a25dfa5d43a8970ea973d2ee4db621b

JesusMcCloud (approved):

nodh:
I'm unhappy with the added dependency on `HttpClient` from  ktor ... is this really necessary?

JesusMcCloud:
how would you make an http request, without an http client?

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
```diff
@@ -386,3 +465,17 @@ class OidcSiopWallet(
     }
 
 }
+
+typealias RequestObjectCandidateRetriever = suspend (Url) -> List<String>
+
+private val HttpClient.asRequestObjectCandidateRetriever: RequestObjectCandidateRetriever
```
nodh, line 471: imho `as` as the prefix indicates a function, but this is a property

acrusage:
> > how would you make an http request, without an http client?
> 
> Why don't we leave that up to the client, i.e. the caller of our library, as it has been the case up until now?

fixed in commit 03a0cdd1303f7c2e0e17cbc4d665635566875f47 
prepared at pull request https://github.com/a-sit-plus/kmm-vc-library/pull/46 along with a fix for https://github.com/a-sit-plus/kmm-vc-library/pull/45#discussion_r1547342169

PR #49, opened 2024-04-09
Improved presentation flexibility

acrusage:
This pull requests adds logic to handle a presentation definition more diligently, but this unfortunately leads to some changes in semantics expected from the "createPresentation" function before that. I hope the changes are reasonable, I tried following the guide from PresentationExchange at:
 - https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-evaluation

I tried to stay conform with openId4vp, but I hope that we don't need to support other presentation specifications?

Semantic changes to be approved:

### Presentation Submissions now only contain matches from input descriptors.
Previously a verifiable presentation contained all the verifiable credentials with the required credential scheme, but not anymore. 
Each input descriptor matches to exactly one verifiable credential, and that verifiable credential must satisfy all constraint fields.
This may actually break some things, as it was previously possible to present multiple AtomicCredentials at once, but after this change each AtomicCredential attribute needs to be requested separately (in separate input descriptors).
-> result.vp.verifiableCredentials now always has size 1!
Impacted Tests: 
- AgentTest."valid presentation is valid": change last "shouldHaveSize" to "1"
- ValidatorVpTest."valid parsed presentation should separate revoked and valid credentials": Change last "shouldHaveSize" to "issuedCredential.successful.size"

### Presentation Submissions where no fields are requested don't return any fields
Previously it seems like not requesting any claims should return all available claims, but not anymore.
Impacted Tests: 
- OidcSiopIsoProtocolTest."test with Fragment for mDL": request an attribute to satisfy the "document.validItems.shouldNotBeEmpty()" assertion
- OidcSiopIsoProtocolTest."test with Fragment for custom attributes": request an attribute to satisfy the "document.validItems.shouldNotBeEmpty()" assertion

### Scope value semantics have somewhat changed
Previously the scope values were used to deduce the requested credential schemes, but not anymore.
As per openid4vp the scope value has the following semantics:

> Such a scope value MUST be an alias for a well-defined Presentation Definition that will be referred to in the presentation_submission response parameter.
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.3-2

scope values are therefore only taken into consideration now in case neither presentationDefinition nor presentationDefinitionUri return a presentation definition.


vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/data/JSONPath.abnf
JesusMcCloud, line 1: this belongs into `src/commonMain/resources`

vclib/src/commonTest/kotlin/at/asitplus/wallet/lib/data/JSONPathParserTest.kt
JesusMcCloud, line 1: don't get me wrong, the JSONPath implementation looks impressive, but we really want a test data set, data-driven testing and property testing covering  random inputs with edge cases

JesusMcCloud (changes requested):
I have yet to reveiw the whole thing, but do try to get some existing test data for the json path impl (see my comment)

acrusage:
Closed due to further work on JSONParser

vclib-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidc/EudiwPidCredentialScheme.kt
nodh, line 1: You could also include https://github.com/a-sit-plus/eu-pid-credential for `commonTest` directly!

PR #53, opened 2024-04-22
 Improved presentation flexibility

acrusage-iaik:
This pull requests adds logic to handle a presentations more diligently, but this unfortunately leads to some changes in semantics expected from the "createPresentation" function before that. I hope the changes are reasonable, I tried following the guide from PresentationExchange at:
https://identity.foundation/presentation-exchange/spec/v2.0.0/#input-evaluation

I tried to stay conform with openId4vp, but I hope that we don't need to support other presentation specifications?

Semantic changes to be approved:
### Presentation Submissions now only contain matches from input descriptors.

Previously a verifiable presentation contained all the verifiable credentials with the required credential scheme, but not anymore.
Each input descriptor matches to exactly one verifiable credential, and that verifiable credential must satisfy all constraint fields.
This may actually break some things, as it was previously possible to present multiple AtomicCredentials at once, but after this change each AtomicCredential attribute needs to be requested separately (in separate input descriptors).
-> result.vp.verifiableCredentials now always has size 1!
Impacted Tests (among others):
- AgentTest."valid presentation is valid": change last "shouldHaveSize" to "1"
- ValidatorVpTest."valid parsed presentation should separate revoked and valid credentials": Change last "shouldHaveSize" to "issuedCredential.successful.size"

### Presentation Submissions where no fields are requested don't return any fields

Previously it seems like not requesting any claims should return all available claims, but not anymore.
Impacted Tests (among others):

- OidcSiopIsoProtocolTest."test with Fragment for mDL": request an attribute to satisfy the "document.validItems.shouldNotBeEmpty()" assertion
- OidcSiopIsoProtocolTest."test with Fragment for custom attributes": request an attribute to satisfy the "document.validItems.shouldNotBeEmpty()" assertion

### Scope value semantics have somewhat changed

Previously the scope values were used to deduce the requested credential schemes, but not anymore.
As per openid4vp the scope values have the following semantics:

    Such a scope value MUST be an alias for a well-defined Presentation Definition that will be referred to in the presentation_submission response parameter.
    https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.3-2

scope values are therefore only taken into consideration now in case neither presentationDefinition nor presentationDefinitionUri return a presentation definition.

### Users can now define a path authorization validator 
This can mainly be done by providing a default authorizer in the constructor of HolderAgent.


vclib-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidc/EudiwPidCredentialScheme.kt
```diff
@@ -7,5 +7,79 @@ object EudiwPidCredentialScheme : ConstantIndex.CredentialScheme {
     override val vcType: String = "EudiwPid1"
     override val isoNamespace: String = "eu.europa.ec.eudiw.pid.1"
     override val isoDocType: String = "eu.europa.ec.eudiw.pid.1"
-    override val claimNames: Collection<String> = listOf("family_name")
+    override val claimNames: Collection<String> = listOf(
```
nodh, line 10: why not use the artifact of https://github.com/a-sit-plus/eu-pid-credential?

vclib-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopIsoProtocolTest.kt
```diff
@@ -99,6 +100,7 @@ class OidcSiopIsoProtocolTest : FreeSpec({
             OidcSiopVerifier.RequestOptions(
                 representation = ConstantIndex.CredentialRepresentation.ISO_MDOC,
                 credentialScheme = ConstantIndex.AtomicAttribute2023,
+                requestedAttributes = listOf("given-name"),
```
nodh, line 103: why not use a constant here?

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
```diff
@@ -87,48 +90,61 @@ interface Holder {
...
      */
     suspend fun createPresentation(
         challenge: String,
         audienceId: String,
-        credentialSchemes: Collection<ConstantIndex.CredentialScheme>? = null,
-        requestedClaims: Collection<String>? = null,
-    ): CreatePresentationResult?
+        presentationDefinition: PresentationDefinition,
+        // TODO: add authorization semantics to detect unauthorized requests
+        // - eg. a service provider asking for an attribute he should not be allowed to see
```
nodh, line 138: imho that's not part of the logic in vclib, especially not at the holder interface

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
```diff
@@ -87,48 +90,61 @@ interface Holder {
...
+    ): KmmResult<HolderResponseParameters>
 
     /**
      * Creates a [VerifiablePresentation] with the given [validCredentials].
      *
      * Note: The caller is responsible that only valid credentials are passed to this function!
      */
     // TODO dont make this public
```
nodh, line 146: is this TODO still relevant?

nodh (changes requested):
Just a few remarks, since this is a draft still.
But wouldn't it be cleaner to create a new, separate, library for JSON Path processing? There seems to be a lot of classes involved.

vclib-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidc/EudiwPidCredentialScheme.kt
acrusage: Should this really be a dependency of the vclib?
What if the EUDIW PID credential changes in the future? This may break the tests, even though the tested functionality still remains correct?
Maybe it would be better to just rename it to TestCredential or something like that?

vclib-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopIsoProtocolTest.kt
acrusage: AtomicAttribute2023 does not specify any attribute constants

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
acrusage: In my opinion it would be actor performing the selective disclosure procedure that must be the one who is aware of the verifiers permissions in accessing them..
Alternatively there would need to be a separate check before or after the selective disclosure has been performed, and in this case the TODO would have to be moved somewhere else?

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
acrusage: I am not sure to be honest...
I think it has already been there before, and I am not aware of its relevance as there is not more information on why the TODO exists in the first place

acrusage:
> Just a few remarks, since this is a draft still. But wouldn't it be cleaner to create a new, separate, library for JSON Path processing? There seems to be a lot of classes involved.

Thanks!
Yes, that would be a better approach, but I'm unfortunately not yet familiar with creating KMP libraries or publishing..

vclib-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidc/EudiwPidCredentialScheme.kt
nodh: Then why not rename it to "TestCredential", or if it tests something EU-PID-specific, move the test to the eu-pid repository? But imho depending on version 1.0.0 of that EU PID artifact isn't that bad, because future updates to that credential should remain backwards-compatible, or be released as 2.0.0, so it won't be picked up by vclib.

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
nodh: You're right, there should be something like a verification method (or function pointer) in the holder class, that Wallets need to implement (with the default implementation to allow all requests).

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
nodh: You're right, that's on my side ...

acrusage-iaik:
@nodh @JesusMcCloud PR is now ready for review & merge

conventions-vclib/src/main/kotlin/VcLibConventions.kt
acrusage-iaik, line 1: Is it normal that these changes are part of a pull request on this repository? Shouldn't this result in changes to the conventions submodule?

vclib-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidc/EudiwPidCredentialScheme.kt
acrusage-iaik: Has been fixed somewhere along the way, not sure exactly where

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
acrusage-iaik: Todo for the future

conventions-vclib/src/main/kotlin/VcLibConventions.kt
JesusMcCloud: I need to evaluate whether we still need the VcLibConventions, now that our generic conventions plugin supports version catalogs. I'm pretty sure that @nodh's suspicion that we should can it is correct. for this PR, though we should leave it like this.
I've assigned myself issue #59 for that.

JesusMcCloud (commented):
externalisation of jsonpath looks good, but is should we not wait for the next release of the library before we continue here? (I'll try to do that today)

acrusage-iaik:
> externalisation of jsonpath looks good, but is should we not wait for the next release of the library before we continue here? (I'll try to do that today)

This should not be an issue for us
The next release only makes changes to function extensions, which we will most likely not use for now

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
nodh: see https://github.com/a-sit-plus/kmm-vc-library/issues/60

vclib-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopInteropTest.kt
```diff
@@ -47,126 +48,108 @@ class OidcSiopInteropTest : FreeSpec({
...
-
         """.trimIndent()
 
         holderSiop = OidcSiopWallet.newInstance(
             holder = holderAgent,
             cryptoService = holderCryptoService,
             remoteResourceRetriever = {
                 if (it == "https://verifier-backend.eudiw.dev/wallet/jarm/" +
-                    "WLFJEn9AGbJfAcEyaQTzzxueqmeRazmsHIkxMRTkGRL1zyI7un-KJWaXtulrfiSS38LlU5ABDB9Zdsfq_11r8Q/jwks.json"
+                    "lif-P02Wm25thTKocReEjQar-KqmmAYMo7xW_nNqTmum6yq0l_1qqLIxn2BYVwKDPU_dd0BGZjN1Cga4kVO_nw/jwks.json"
```
nodh, line 143: Why did this change?

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/HolderAgent.kt
```diff
@@ -139,88 +189,204 @@ class HolderAgent(
...
+            // prefer iso credentials and sd jwt credentials over plain vc credentials
+            // -> they support selective disclosure!
+            when (it) {
+                is SubjectCredentialStore.StoreEntry.Vc -> 2
+                is SubjectCredentialStore.StoreEntry.SdJwt -> 1
+                is SubjectCredentialStore.StoreEntry.Iso -> 1
+            }
+        } ?: return KmmResult.failure(CredentialRetrievalException())
+
+        data class CandidateInputMatchContainer(
```
nodh, line 235: I would not put a `data class` inside a method. Maybe move out some parts of this large method?

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/HolderAgent.kt
```diff
@@ -292,6 +471,102 @@ class HolderAgent(
         }
         return Holder.CreatePresentationResult.Signed(jws.serialize())
     }
+}
+
+// in openid4vp, the claims to be presented are described using a JSONPath, so compiling this to a JsonElement seems reasonable
+fun SubjectCredentialStore.StoreEntry.toJsonElement(): JsonElement {
```
nodh, line 477: maybe put that into a separate file, so that this file here doesn't get over 500 lines long (which is way too much)

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/data/dif/InputEvaluator.kt
```diff
@@ -0,0 +1,147 @@
...
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonNull
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.booleanOrNull
+import kotlinx.serialization.json.doubleOrNull
+import kotlinx.serialization.json.longOrNull
+
+/*
```
nodh, line 16: Please use `/**` and also move the comment in `//` into here

vclib/src/commonMain/resources/grammar/JsonPath.abnf
```diff
@@ -0,0 +1,149 @@
+; specification: https://datatracker.ietf.org/doc/rfc9535/
```
nodh, line 1: isn't this file part of the JSONPath library now? Same for the `g4` ones

vclib/src/commonTest/kotlin/at/asitplus/wallet/lib/agent/AgentSdJwtTest.kt
```diff
@@ -38,10 +45,29 @@ class AgentSdJwtTest : FreeSpec({
...
     "simple walk-through success" {
         issueDummyCredentials(holder, issuer, holderCryptoService)
-        val vp =
-            holder.createPresentation(challenge, verifier.identifier, requestedClaims = listOf("given-name")).also {
-                it.shouldNotBeNull()
-            }
+        val presentationParameters =
+            holder.createPresentation(
+                challenge,
+                verifier.identifier,
```
nodh, line 51: Code here seems duplicated with functions below

nodh (changes requested):
Great work, just some minor "editorial" changes requested

vclib-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopInteropTest.kt
acrusage-iaik: I just did a new presentation and copied the new parameters there

It seems like their presentation request changed from back then to now, so I figured it would be a good idea to update this

JesusMcCloud (approved):
approved, pending the change requests by @nodh

vclib/src/commonTest/kotlin/at/asitplus/wallet/lib/agent/AgentSdJwtTest.kt
acrusage-iaik: fixed in commit 950b306a68c5b7a9af3591d61dead5fe364bd94c

vclib/src/commonMain/resources/grammar/JsonPath.abnf
acrusage-iaik: fixed in 3e4c8cd54334e7b7ed0bb56817ae64d8ed0d8d0d

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/data/dif/InputEvaluator.kt
acrusage-iaik: fixed in 3e4c8cd54334e7b7ed0bb56817ae64d8ed0d8d0d

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/HolderAgent.kt
acrusage-iaik: fixed in 3e4c8cd54334e7b7ed0bb56817ae64d8ed0d8d0d

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/HolderAgent.kt
acrusage-iaik: Moved to `vclib/.../data/CredentialToJsonConverter` in commit 3e4c8cd54334e7b7ed0bb56817ae64d8ed0d8d0d

After refactoring for comment https://github.com/a-sit-plus/kmm-vc-library/pull/53#discussion_r1593812217 it got way too big though, so I also moved the `create*Presentation-files` to their own class: `vclib/.../agent/VerifiablePresentationFactory`
I'm not so sure whether this makes the code too loose and hard to grasp though, and I'm also unsure about the location and the name of the new class

acrusage-iaik:
> Great work, just some minor "editorial" changes requested

Unfortunately I also implemented other things while doing the refactoring to make the method smaller, so I guess I'll have to re-request a review..
Sorry about that!

JesusMcCloud:
To be frank: I lack the context to judge these last changes.

nodh (changes requested):
Great work! Some minor questions remain

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
```diff
@@ -128,14 +129,18 @@ interface Holder {
 
     /**
      * Creates an array of [VerifiablePresentation] and a [PresentationSubmission] to match
-     * the [presentationDefinition].
+     * the [presentationOption].
+     *
+     * [presentationOption] - run DescriptorCombination.findValidDescriptorCombinations and
```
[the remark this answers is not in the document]
acrusage-iaik: fixed/reworked in commit 79ba0e4f03b62d8f3e5fd2fca38190a3b113e2a9

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
```diff
@@ -63,7 +61,12 @@ class OidcSiopWallet(
      * Need to verify the request object serialized as a JWS,
      * which may be signed with a pre-registered key (see [OpenIdConstants.ClientIdSchemes.PRE_REGISTERED]).
      */
-    private val requestObjectJwsVerifier: RequestObjectJwsVerifier = RequestObjectJwsVerifier { _, _ -> true }
+    private val requestObjectJwsVerifier: RequestObjectJwsVerifier = RequestObjectJwsVerifier { _, _ -> true },
+    /**
+     * Need to implement if the presentation definition needs to be derived from a scope value.
+     * See [ScopePresentationDefinitionRetriever] for implementation instructions.
+     */
+    private val scopePresentationDefinitionRetriever: ScopePresentationDefinitionRetriever? = null,
```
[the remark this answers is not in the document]
acrusage-iaik: added tests in commit e73130c89d6226fd87bb80c5a83d755d9792b851

acrusage-iaik:
> To be frank: I lack the context to judge these last changes.

I removed work towards submission requirements as it wasn't polished at all, maybe it's easier to review now?

nodh (approved):
Let's merge everything together for vclib 3.7.0 and see how it works out :-)

JesusMcCloud:
Well now that its out, we could just bump jsonpath to 2.0.0?

PR #61, opened 2024-05-08
Add OAuth2ServerMetadata

n0900:
Implements RFC8414

### DIFF
--- a/vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/OAuth2AuthorizationServer.kt
+++ b/vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/OAuth2AuthorizationServer.kt
@@ -30,6 +30,6 @@ interface OAuth2AuthorizationServer {
     /**
      * Provide necessary [OAuth2AuthorizationServerMetadata] JSON for a client to be able to authenticate
      */
-    suspend fun provideMetadata(): String
+    suspend fun provideMetadata(): KmmResult<OAuth2AuthorizationServerMetadata>
 }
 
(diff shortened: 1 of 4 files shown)


nodh (approved):

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/OAuth2AuthorizationServer.kt
```diff
@@ -26,5 +26,10 @@ interface OAuth2AuthorizationServer {
 
     // TODO How is this supposed to happen when using an external Authorization Server?
     suspend fun verifyAndRemoveClientNonce(nonce: String): Boolean
+
+    /**
+     * Provide necessary [OAuth2AuthorizationServerMetadata] JSON for a client to be able to authenticate
+     */
+    suspend fun provideMetadata(): String
```
nodh, line 33: Why not `KmmResult<OAuth2AuthorizationServerMetadata>`?

n0900: According to the spec it should return json encoded metadata. Wouldnt KmmResult interfere with the content-negotiation between client-server downstream?

nodh: In a remote setting, the implementation would fetch the metadata from the remote service, decode it and pass it on (or an failure wrapped in a `KmmResult`).
Locally, passing the object makes even more sense.

nodh (approved):

PR #64, opened 2024-05-14
Bump: Dependencies

acrusage-iaik:
Bumping dependency version of: 

- jsonpath

Issue first observed in: https://github.com/a-sit-plus/kmm-vc-library/pull/53#issuecomment-2109983034


JesusMcCloud (approved):
ship it!

PR #66, opened 2024-05-14
Feature: Expose input descriptor matching to library users

acrusage-iaik:
This will allow the wallet to delegate the evaluation of what is to be presented to the vclib

### DIFF
--- a/vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/HolderAgent.kt
+++ b/vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/HolderAgent.kt
@@ -324,7 +324,7 @@ class HolderAgent(
     ): Map<InputDescriptor, CandidateInputMatchContainer?> {
         return inputDescriptors.associateWith { inputDescriptor ->
             credentials.firstNotNullOfOrNull { credential ->
-                evaluateInputDescritorAgainstCredential(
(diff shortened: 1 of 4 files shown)


nodh (approved):

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/HolderAgent.kt
```diff
@@ -302,20 +333,20 @@ class HolderAgent(
...
                         credential = credential,
+                        fieldQueryResults = it,
                     )
                 }
             }
         }
     }
 
-    private fun findDescriptorInputMatching(
+    private fun evaluateInputDescritorAgainstCredential(
```
JesusMcCloud, line 344: typo

JesusMcCloud (changes requested):
Fix the typo and you're good to go!

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/HolderAgent.kt
acrusage-iaik: fixed in commit f85a5a54cacd343dd4bbdf10a5fbcd05dfb07d03

JesusMcCloud (approved):

PR #78, opened 2024-05-29
Update SD-JWT, Support encryption of SIOP responses

nodh:


JesusMcCloud (approved):
Why can't I shake the feeling, I've been here and seen this before?!

nodh:
> Why can't I shake the feeling, I've been here and seen this before?!

Maybe you're thinking of https://github.com/a-sit-plus/kmm-vc-library/pull/62

PR #89, opened 2024-06-20
Use catching scope function to reduce verbosity

nodh:


JesusMcCloud (approved):
that's a no-brainer, awesome!

PR #97, opened 2024-07-09
Minimize usage of identifiers

nodh:
Based on https://github.com/a-sit-plus/kmp-crypto/pull/95 remove use of `JsonWebKey.identifier`


vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Validator.kt
```diff
@@ -442,3 +437,12 @@ class Validator(
     }
 
 }
+
+private fun JsonWebKey.equalsCryptographically(second: JsonWebKey) =
```
JesusMcCloud, line 441: That should be part of kmp-crypto imho

JesusMcCloud (approved):
image

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Validator.kt
nodh: See https://github.com/a-sit-plus/kmp-crypto/pull/96

vclib-aries/src/commonMain/kotlin/at/asitplus/wallet/lib/aries/PresentProofProtocol.kt
n0900, line 1: senderKey.didEncoded only fails in case it is an invalid jwk/ a key format we do not support. Potentially it would be better to throw in this case?

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
n0900, line 1: it.didEncoded only fails in case it is an invalid jwk/ a key format we do not support. Potentially it would be better to throw in this case?

nodh: Since this is only a fallback (as usually the `options` are set), I would not throw in that case.

vclib-aries/src/commonMain/kotlin/at/asitplus/wallet/lib/aries/PresentProofProtocol.kt
nodh: Since this is only a fallback (as usually the `options` are set), I would not throw in that case.

vclib-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopWallet.kt
nodh: I wouldn't throw in that case, because this is just to extract the correct audience string ... unsupported keys should either not matter or be handled somewhere else.

n0900 (approved):

PR #100, opened 2024-07-10
Add credential matching functionality

n0900:
Admittetly even more permissive than before this function now also returns the associated representation and thus extending its use cases.

### DIFF
--- a/vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/data/AttributeIndex.kt
+++ b/vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/data/AttributeIndex.kt
@@ -52,12 +52,14 @@ object AttributeIndex {
             .firstOrNull { it.isoDocType!!.startsWith(docType) || docType.startsWith(it.isoDocType!!) }
     }
 
-    // TODO Probably shouldn't be that permissive
-    fun resolveCredentialScheme(attributeType: String): ConstantIndex.CredentialScheme? {
-        return resolveAttributeType(attributeType)
-            ?: resolveSdJwtAttributeType(attributeType)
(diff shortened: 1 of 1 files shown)


vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/data/AttributeIndex.kt
```diff
@@ -52,12 +52,14 @@ object AttributeIndex {
...
-    // TODO Probably shouldn't be that permissive
-    fun resolveCredentialScheme(attributeType: String): ConstantIndex.CredentialScheme? {
-        return resolveAttributeType(attributeType)
-            ?: resolveSdJwtAttributeType(attributeType)
-            ?: resolveIsoNamespace(attributeType)
-            ?: resolveSchemaUri(attributeType)
+    /**
+     * Compares the input to all CredentialScheme identifiers and on match returns it plus its associated `ConstantIndex.CredentialRepresentation` if applicable
+     */
+    fun resolveCredential(input: String): Pair<ConstantIndex.CredentialScheme, ConstantIndex.CredentialRepresentation?>? {
```
nodh, line 58: Where would we need such a method? Doesn't the caller know if its an SD-JWT or ISO-Credential?

nodh (changes requested):

vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/data/AttributeIndex.kt
n0900: The caller would know, however a receiver of an identifier would not. To identify it using VClib one needs to repeat exactly these steps. We currently have no way of matching an identifier to a representation.

PR #103, opened 2024-07-29
Rebranding

JesusMcCloud:
This should go through first.
 1. Then publish snapshot to sonatype (without relocation POM, so we don't run into the KLIB bug.)
 2. Then publish snapshots of EU PID and MDL credentials to sonatype, because we require them here for testing
 4. Then publish this rebranded library with relocation POM to sonatype
 5. Then update all credentials (refactor `initWithVcLib` to `initWithVck` or whatever is preferred
 6. Update credentials collection and a-sit-plus.github.io
 7. Update everything else that depends on this work

After we know all this went well, we should do another round, using the repranded kmp-crypto and its verifier and the rebranded jsonpath library.


.github/workflows/build-jvm.yml
```diff
@@ -14,21 +14,21 @@ jobs:
...
+          name: ckib
           path: |
-            vclib/build/libs/*jar
-      - name: Upload jar vclib-aries
+            ckib/build/libs/*jar
+      - name: Upload jar ckib-aries
         uses: actions/upload-artifact@v3
         with:
-          name: vclib-aries
+          name: ckib-aries
```
nodh, line 26: typos everywhere?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/data/Json.kt
```diff
@@ -27,11 +27,7 @@ internal object JsonCredentialSerializer {
 }
 
 val jsonSerializer by lazy {
-    Json {
-        prettyPrint = false
-        encodeDefaults = false
-        classDiscriminator = "type"
-        ignoreUnknownKeys = true
+    Json(from = at.asitplus.crypto.datatypes.jws.io.jsonSerializer) {
```
nodh, line 30: I would not remove the explicit configuration of the serializer in this library

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/iso/cborSerializer.kt
```diff
@@ -43,11 +43,10 @@ internal object CborCredentialSerializer {
 }
 
 @OptIn(ExperimentalSerializationApi::class)
-val cborSerializer by lazy {
-    Cbor {
+val vckCborSerializer by lazy {
+    Cbor(from = at.asitplus.crypto.datatypes.cose.io.cborSerializer) {
```
nodh, line 47: I would not remove the explicit configuraiton of the serializer in this library

nodh (changes requested):

.github/workflows/build-jvm.yml
JesusMcCloud: what in the…? i could swear I used the same search&replace expressions as in the other files

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/data/Json.kt
JesusMcCloud: Makes sense fro JSON

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/iso/cborSerializer.kt
JesusMcCloud: Here I disagree, since this one is a COSE-complaint serializer and this is exactly what we want here. Everything beyond COSE compliance is explicitly configured, just as it was before.
What it does lack is a proper API documentation in kmp-crypto

JesusMcCloud:
same procedure as with 4.0.0: includes dependencies and publishes them to mavenLocal, to work around kotlin build bugs, the last one of which will be gone for good with the next serializationr release end of august/beginning of september, once kotlin 2.0.20 is out.

Porposed course of action:
* @nodh review this
* @JesusMcCloud publish signum and jsonPath to maven central
* @JesusMcCloud remove all the included stuff
* @JesusMcCloud publish this one here to maven central
* @JesusMcCloud republish all credentials
* @JesusMcCloud update all services and apps to use the new coordinates

nodh (approved):

PR #107, opened 2024-08-12
Feature/request option refactor

n0900:
Merges `OidcSiopVerifier.RequestOptions` and `WalletService.RequestOptions` into a single class and into a seperate file which also contains all helper functions which were used to deal with manipulating RequestOptions only.

Furthermore, after looking through https://identity.foundation/presentation-exchange/spec/v1.0.0/#input-descriptor-object again I found that `InputDescriptor.schema` is actually MUST and thus can be used to get our credential scheme way easier than previously. Maybe I have misread - please double check. This breaks two interop tests, namely 'EUDI AuthnRequest can be parsed' and 'Idemia Interop Request'.


JesusMcCloud:
given the interop tests are the failing ones, I'm pretty sure we made  `InputDescriptor.schema` optional to successfully interop with other parties. Still, eve if this is the case, we should explicitly indicate it in the code

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/RequestOptions.kt
```diff
@@ -0,0 +1,171 @@
...
+) {
+    internal fun buildScope() = listOfNotNull(
+        SCOPE_OPENID,
+        SCOPE_PROFILE,
+        credentialScheme?.sdJwtType,
+        credentialScheme?.vcType,
+        credentialScheme?.isoNamespace
+    ).joinToString(" ")
+
+    internal fun setRedirectUrl(): Boolean =
```
JesusMcCloud, line 63: that's a misnomer! reads like a setter

JesusMcCloud (changes requested):
renaming of method required

n0900:
> given the interop tests are the failing ones, I'm pretty sure we made `InputDescriptor.schema` optional to successfully interop with other parties. Still, eve if this is the case, we should explicitly indicate it in the code

I want to add that if we decide to keep the parameter nullable for interop, we can add an Exception to `RequestOptions.fromInputParameters` so that it still works. So either choice is fine.

n0900:
In order to unify `OidcSiopVerifier.RequestOptions` and `WalletService.RequestOptions` I added clock and set state to not nullable. This is not a real change since RequestOptions is an internal class and by default it got a value automatically - null has never been used.

JesusMcCloud:
The way I read this, it seems way cleaner and makes a lot of sense. Regarding interop: Maybe warn, default to the PID schema from the implementing acts and hope for the best is a valid strategy

.github/workflows/cla.yml
JesusMcCloud, line 1: just to add context for @nodh: this was required to not make github stuck, even though it is run on the target branch. makes no sense to me, but that's the way it is

JesusMcCloud (approved):
Fine for me, but I am pretty sure @nodh will tell us that the property was nullable on prupose to make interop tests green, so we he will probably request changes. The fix will probably be defaulting to EU PID if it is indeed null. Of course, this requires thorough documentation

CHANGELOG.md
```diff
@@ -1,6 +1,16 @@
 # Changelog
 
 Release NEXT:
+* Refactor `OidcSiopVerifier.RequestOptions` and `WalletService.RequestOptions` into one separate class `RequestOptions`
+  * Adds `InputParameter.toRequestOptions()` functionality
+
+Release 4.1.1 (Bugfix Release):
```
nodh, line 7: But 4.1.1 is now duplicated in the changelog?

gradle.properties
```diff
@@ -12,7 +12,7 @@ kotlin.mpp.enableCInteropCommonization=true
 kotlin.mpp.stability.nowarn=true
 kotlin.native.ignoreDisabledTargets=true
 
-artifactVersion = 4.2.0-SNAPSHOT
+artifactVersion = 4.1.1
```
nodh, line 15: Why is this 4.1.1 on a feature branch?

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt
```diff
@@ -0,0 +1,236 @@
...
+    path = listOf(
+        NormalizedJsonPath(
+            NormalizedJsonPathSegment.NameSegment(this?.isoNamespace ?: "mdoc"),
+            NormalizedJsonPathSegment.NameSegment(attributeType),
+        ).toString()
+    ), intentToRetain = false
+)
+
+private fun ConstantIndex.CredentialScheme.toVcConstraint() = if (supportsVcJwt) ConstraintField(
+    path = listOf("$.type"), filter = ConstraintFilter(
```
nodh, line 144: Indendation seems off

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt
```diff
@@ -0,0 +1,236 @@
...
+
+/**
+ * Miscellaneous helper functions regarding [ConstraintField]
+ */
+private fun Collection<String>.createConstraints(
+    credentialRepresentation: CredentialRepresentation,
+    credentialScheme: ConstantIndex.CredentialScheme?,
+): Collection<ConstraintField> = map {
+    if (credentialRepresentation == ISO_MDOC) credentialScheme.toConstraintField(
+        it
```
nodh, line 129: Indendation seems off

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt
```diff
@@ -0,0 +1,236 @@
...
+private fun ConstantIndex.CredentialScheme.toVcConstraint() = if (supportsVcJwt) ConstraintField(
+    path = listOf("$.type"), filter = ConstraintFilter(
+        type = "string",
+        pattern = vcType,
+    )
+) else null
+
+private fun ConstantIndex.CredentialScheme.toSdJwtConstraint() = if (supportsSdJwt) ConstraintField(
+    path = listOf("$.vct"), filter = ConstraintFilter(
+        type = "string", pattern = sdJwtType!!
```
nodh, line 152: Indendation seems off

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt
```diff
@@ -0,0 +1,236 @@
...
+import at.asitplus.wallet.lib.oidvci.CredentialRequestProof
+import at.asitplus.wallet.lib.oidvci.SupportedCredentialFormatDefinition
+import at.asitplus.wallet.lib.oidvci.toRequestedClaimsIso
+import at.asitplus.wallet.lib.oidvci.toRequestedClaimsSdJwt
+import com.benasher44.uuid.uuid4
+import io.ktor.http.*
+import kotlinx.datetime.Clock
+
+
+data class RequestOptions(
```
nodh, line 37: But now not all parameters here make sense for both, SIOPv2 and OID4VCI, right?

nodh:
> Fine for me, but I am pretty sure @nodh will tell us that the property was nullable on prupose to make interop tests green, so we he will probably request changes. The fix will probably be defaulting to EU PID if it is indeed null. Of course, this requires thorough documentation

Indeed, even now, the EUDIW verifier at <https://verifier.eudiw.dev/pid-full> generates authn requests that do not contain a `schema`, i.e. click on requesting specific attributes, and expend the presentation definition to

```
{
	"id": "33ac795e-599b-460c-ab3d-db4c1ccf73fd",
	"input_descriptors": [
		{
			"id": "eu.europa.ec.eudi.pid.1",
			"name": "EUDI PID",
			"purpose": "We need to verify your identity",
			"format": {
				"mso_mdoc": {
					"alg": [
						"ES256",
						"ES384",
						"ES512"
					]
				}
			},
			"constraints": {
				"fields": [
					{
						"path": [
							"$['eu.europa.ec.eudi.pid.1']['age_over_18']"
						],
						"intent_to_retain": false
					}
				]
			}
		}
	]
}
```

But we should also not default to the PID scheme (which we can't, because that's an external artifact?)

nodh (changes requested):

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/data/dif/InputDescriptor.kt
```diff
@@ -21,7 +21,7 @@ data class InputDescriptor(
     @SerialName("format")
     val format: FormatHolder? = null,
     @SerialName("schema")
-    val schema: Collection<SchemaReference>? = null,
+    val schema: Collection<SchemaReference>,
```
nodh, line 24: So (see comment) better leave it nullable?

gradle.properties
n0900: merging error. mb

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt
n0900: OID4VCI does not use `responseMode`, `clientMetadataUrl` and `encryption`, however by their default values they do not interfere with this use case (nullable) and they are ignored for all OID4VCI use-cases.
The same goes for SIOPv2 and clock. Since we never use RequestOptions as object but rather access it's members directly this does not influence the usage. Additionally we can make clock transient for serialization to make sure that it is never leaked.

n0900: Should i add this to the variable documentation?

JesusMcCloud:
> > Fine for me, but I am pretty sure @nodh will tell us that the property was nullable on prupose to make interop tests green, so we he will probably request changes. The fix will probably be defaulting to EU PID if it is indeed null. Of course, this requires thorough documentation
> 
> Indeed, even now, the EUDIW verifier at https://verifier.eudiw.dev/pid-full generates authn requests that do not contain a `schema`, i.e. click on requesting specific attributes, and expand the presentation definition to
> 
> ```
> {
> 	"id": "33ac795e-599b-460c-ab3d-db4c1ccf73fd",
> 	"input_descriptors": [
> 		{
> 			"id": "eu.europa.ec.eudi.pid.1",
> 			"name": "EUDI PID",
> 			"purpose": "We need to verify your identity",
> 			"format": {
> 				"mso_mdoc": {
> 					"alg": [
> 						"ES256",
> 						"ES384",
> 						"ES512"
> 					]
> 				}
> 			},
> 			"constraints": {
> 				"fields": [
> 					{
> 						"path": [
> 							"$['eu.europa.ec.eudi.pid.1']['age_over_18']"
> 						],
> 						"intent_to_retain": false
> 					}
> 				]
> 			}
> 		}
> 	]
> }
> ```
> 
> But we should also not default to the PID scheme (which we can't, because that's an external artifact?)

I think we need to make a general decision: do we want to be lenient or spec-conforming. This particular issue should be reported at the eudiw verifier. The simple fact that having this property simplifies processing considerably is reason enough to think that there's a very good reason why the spec demands it (assuming we did not misinterpret it).  
Put differently:

> Somebody is wrong on the Internet!

n0900:
> > > Fine for me, but I am pretty sure @nodh will tell us that the property was nullable on prupose to make interop tests green, so we he will probably request changes. The fix will probably be defaulting to EU PID if it is indeed null. Of course, this requires thorough documentation
> > 
> > 
> > Indeed, even now, the EUDIW verifier at https://verifier.eudiw.dev/pid-full generates authn requests that do not contain a `schema`, i.e. click on requesting specific attributes, and expand the presentation definition to
> > ```
> > {
> > 	"id": "33ac795e-599b-460c-ab3d-db4c1ccf73fd",
> > 	"input_descriptors": [
> > 		{
> > 			"id": "eu.europa.ec.eudi.pid.1",
> > 			"name": "EUDI PID",
> > 			"purpose": "We need to verify your identity",
> > 			"format": {
> > 				"mso_mdoc": {
> > 					"alg": [
> > 						"ES256",
> > 						"ES384",
> > 						"ES512"
> > 					]
> > 				}
> > 			},
> > 			"constraints": {
> > 				"fields": [
> > 					{
> > 						"path": [
> > 							"$['eu.europa.ec.eudi.pid.1']['age_over_18']"
> > 						],
> > 						"intent_to_retain": false
> > 					}
> > 				]
> > 			}
> > 		}
> > 	]
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But we should also not default to the PID scheme (which we can't, because that's an external artifact?)
> 
> I think we need to make a general decision: do we want to be lenient or spec-conforming. This particular issue should be reported at the eudiw verifier. The simple fact that having this property simplifies processing considerably is reason enough to think that there's a very good reason why the spec demands it (assuming we did not misinterpret it). Put differently:
> 
> > Somebody is wrong on the Internet!

It turns out I was wrong on the internet since since we cannot use schema-matching according to [v2.0.0 or later](https://identity.foundation/presentation-exchange/spec/v2.1.1/#input-descriptor). We need to update our references to the presentation exchange spec

PR #109, opened 2024-08-19
Add InputDescriptor.toRequestOptions

n0900:
Adds inverse operations to get back RequestOptions which were used to generate InputDescriptor.
This is useful if an incoming request needs to be manipulated.

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt
@@ -121,6 +121,12 @@ data class RequestOptions(
 /**
  * Miscellaneous helper functions regarding [ConstraintField]
  */
+private fun Collection<ConstraintField>.toRequestedAttributes(): List<String> {
+    val regex = "[a-zA-Z0-9_-]+".toRegex()
+    val rawAttributes = this.map { constraint -> constraint.path.last().split("[").last() }
+    return rawAttributes.map { regex.find(it)?.value ?: "" }
+}
+
 private fun Collection<String>.createConstraints(
     credentialRepresentation: CredentialRepresentation,
     credentialScheme: ConstantIndex.CredentialScheme?,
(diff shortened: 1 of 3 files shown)


vck-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/openid/RequestOptionUnitTest.kt
```diff
@@ -0,0 +1,38 @@
...
+    val representations = listOf(
+        ConstantIndex.CredentialRepresentation.PLAIN_JWT,
+        ConstantIndex.CredentialRepresentation.SD_JWT,
+        ConstantIndex.CredentialRepresentation.ISO_MDOC
+    )
+
+
+    representations.forEach { representation ->
+        "${representation.name}: RequestOptions -> InputDescriptor -> RequestOptions quasi equality" {
+            repeat(3) {
```
nodh, line 19: Why repeat it three times?

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt
```diff
@@ -233,4 +239,35 @@ private fun CredentialRepresentation.toFormat() = when (this) {
     PLAIN_JWT -> CredentialFormatEnum.JWT_VC
     SD_JWT -> CredentialFormatEnum.VC_SD_JWT
     ISO_MDOC -> CredentialFormatEnum.MSO_MDOC
+}
+
+/**
+ * Inverse operation of [RequestOptions.toInputDescriptor]
```
nodh, line 245: just add a little more documentation, i.e. when this would be needed?

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt
```diff
@@ -121,6 +121,12 @@ data class RequestOptions(
 /**
  * Miscellaneous helper functions regarding [ConstraintField]
  */
+private fun Collection<ConstraintField>.toRequestedAttributes(): List<String> {
+    val regex = "[a-zA-Z0-9_-]+".toRegex()
+    val rawAttributes = this.map { constraint -> constraint.path.last().split("[").last() }
```
nodh, line 126: This seems it should be handled by JSON Path? / cc @acrusage-iaik

nodh (changes requested):

vck-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/openid/RequestOptionUnitTest.kt
n0900: To get more permutations of requested attributes i.e. empty, all etc

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt
n0900: since we cannot use schema-matching according to [v2.0.0 or later](https://identity.foundation/presentation-exchange/spec/v2.1.1/#input-descriptor) this needs to be reworked.

PR #115, opened 2024-08-28
Extract data classes

nodh:
Move data classes for DIF and OpenID to separate modules, i.e. decoupling them from the vck logic.


nodh:
@JesusMcCloud  Please have a thorough look at the gradle files, because I'm sure there are some shorthand functions that I was not aware of for all those dependency declarations.

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/Json.kt
n0900, line 1: Remove since it is replaced by `vckJsonSerializer`?

n0900 (changes requested):

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/Json.kt
nodh: Can't remove, because that serializer here is used by the data classes

n0900 (approved):

JesusMcCloud:
I need to double-check, but I'm pretty sure most of the dependencies can go away, as they are transitive dependencies of signum. Tomorrow afternoon is a good time for this

JesusMcCloud:
let's keep it that way. I just created #117 to remind me of this, once everything is merged.

PR #125, opened 2024-09-11
OAuth DPoP

nodh:
I'm also thinking about extracting that token functionality from `WalletService`, but this would be better suited in a future PR.

### DIFF
--- a/.github/workflows/test-jvm.yml
+++ b/.github/workflows/test-jvm.yml
@@ -13,7 +13,7 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
       - name: Run tests
-        run: ./gradlew -Dpublishing.excludeIncludedBuilds=true jvmTest
+        run: ./gradlew jvmTest
       - name: Test Report
         uses: dorny/test-reporter@v1
         if: success() || failure()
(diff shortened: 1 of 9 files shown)


n0900 (approved):

JesusMcCloud (approved):
lgtm

nodh:
> Jvm Build fails but tests succeed? Implementation looks good to me.

` e: file:///home/runner/work/vck/vck/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/WalletService.kt:738:9 No parameter with name 'httpMethod' found.`

@JesusMcCloud shouldn't this be a composite build, with the checked-in code from the submodule?

JesusMcCloud:
So apparently, the jvmTest action [disables composite builds](https://github.com/a-sit-plus/vck/actions/runs/10881884620/workflow#L16). I've now reverted this change, BUT we need to have a strategy for testing releases. My proposal: Introduce an `allTests` action that disables composite builds, which will run only on `release/xxx` branches an `main`. D'Accord?!

FYI: snapshot publishing works now for Signum, so we can depend on the actual version catalogs of the signum commit included in the composite build. We should, however, pin this snapshot version to a specific build (e.g. `4.0.0-20240917.201628-3` for indispensable and `0.3.0-20240917.201628-6` for Supreme. if you agree, just assign me an issue with reference to this comment.

PR #130, opened 2024-09-19
OpenId4VCI Draft 14

nodh:
Update our implementation of https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html to Draft 14

PR #127, opened 2024-09-16
Feature/signature services

n0900:
Adds remaining necessary data classes for the proposed UC5 remote signature flow.

This should finish the implementation of the OAuth2/CSC portion of the process.
Does not (yet) include OIDVCI extensions. Does not include any `credential/info` data classes or logic defined by CSC
Helper classes also still rough.

PR #131, opened 2024-09-19
Extract OAuth 2.0 functionality

nodh:
While working on https://github.com/a-sit-plus/vck/pull/130 it appeared to me we should extract more logic from OAuth 2.0 to decouple it from issuing, which should help with the ongoing work in https://github.com/a-sit-plus/vck/pull/127

### DIFF
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/SupportedCredentialFormatDefinition.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/SupportedCredentialFormatDefinition.kt
@@ -15,7 +15,7 @@ data class SupportedCredentialFormatDefinition(
      * according to (VC_DATA), Section 4.3, e.g. `VerifiableCredential`, `UniversityDegreeCredential`
      */
     @SerialName("type")
-    val types: Collection<String>? = null,
+    val types: Set<String>? = null,
 
     /**
      * OID4VCI:
@@ -26,7 +26,4 @@ data class SupportedCredentialFormatDefinition(
     @SerialName("credentialSubject")
     val credentialSubject: Map<String, CredentialSubjectMetadataSingle>? = null,
 
-    // TODO is present in EUDIW issuer ... but is this really valid?
-    @SerialName("claims")
-    val claims: Map<String, RequestedCredentialClaimSpecification>? = null,
 )
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/OAuth2Client.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/OAuth2Client.kt
@@ -84,11 +84,12 @@ class OAuth2Client(
         data class Code(val code: String) : AuthorizationForToken()
 
         /**
-         * Pre-auth code from [CredentialOfferGrants.preAuthorizedCode] in [CredentialOffer.grants],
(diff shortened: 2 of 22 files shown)


openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/SupportedCredentialFormatDefinition.kt
```diff
@@ -15,7 +15,7 @@ data class SupportedCredentialFormatDefinition(
      * according to (VC_DATA), Section 4.3, e.g. `VerifiableCredential`, `UniversityDegreeCredential`
      */
     @SerialName("type")
-    val types: Collection<String>? = null,
+    val types: Set<String>? = null,
```
JesusMcCloud, line 18: It just hit me: A set is unordered by definition. I know I voted for replacing lists with sets in the past and we made those changes, but… how do we guarantee serialization will be canonical? This is a larger issue. The easiest fix for this issue would be (globally) to use a sorted set, or a set that guarantees to preserve the order of elements. Could be that I am missing something here, but in case I am not, please assign an issue to me and I'll create a PR to fix this globally.

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/OAuth2Client.kt
```diff
@@ -0,0 +1,174 @@
...
+     * }
+     * val authn = AuthenticationResponseParameters.deserialize(authnResponse.bodyAsText()).getOrThrow()
+     * ```
+     *
+     * @param state to keep internal state in further requests
+     * @param scope in OID4VCI flows the value `scope` from [IssuerMetadata.supportedCredentialConfigurations]
+     * @param authorizationDetails from RFC 9396 OAuth 2.0 Rich Authorization Requests
+     * @param resource from RFC 8707 Resource Indicators for OAuth 2.0, in OID4VCI flows the value
+     * of [IssuerMetadata.credentialIssuer]
+     */
```
JesusMcCloud, line 55: the code sample is very nice, but don't we already have some ktor dependency in this project, so we could just add the `authnRequest.encodeToParameters().forEach { parameters.append(it.key, it.value) }` as a ready-made extension function in this PR? This obviously affects the other comments too

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/SupportedCredentialFormatDefinition.kt
nodh: I don't think there is ever an issue with canonicalization in serialization of OAuth structures ... but maybe I just don't see it?!

n0900 (approved):

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/SupportedCredentialFormatDefinition.kt
JesusMcCloud: I was more thinking about `SdJwtSigned` and other data that is serialized and sqigned. Signum might be affected too. I'm filing an issue and assigning you.

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/OAuth2Client.kt
JesusMcCloud: nvm. there is no such dependency. mixed up projects

PR #133, opened 2024-09-19
Fix/dependency cleanup

### LINKED ISSUE #117: Cleanup dependency mess (See #115)

JesusMcCloud:
fixes #117 (requires signum update only for new signature of conventions)


JesusMcCloud:
(also: no changelog, because this is just internal and should have absolutely zero effect to outside consumers of this lib)

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/KeyMaterial.kt
```diff
@@ -12,6 +12,7 @@ import at.asitplus.signum.supreme.asKmmResult
 import at.asitplus.signum.supreme.sign.EphemeralKey
 import at.asitplus.signum.supreme.sign.Signer
 import io.github.aakira.napier.Napier
+import kotlinx.coroutines.Job
```
nodh, line 15: shouldn't be here

nodh (approved):

PR #139, opened 2024-10-01
Release 5.0.0

nodh:


JesusMcCloud (approved):
So this goes into develop, published as snapshot, then come the credentials, and then it will be merged into main and released without the SNAPSHOT suffix?

PR #142, opened 2024-10-07
Fix decoding ISO structures without value tags

nodh:


JesusMcCloud (commented):
Just so that I get this right: we are now very lenient, and don't care at all for tags?1
Would it not be better to try and parse with tag verification. if that fails, try again without tag verification and warn, if this succeeds? I mean: this is transitional, is it not? or do we want to permanently remain lenient?

nodh:
> Just so that I get this right: we are now very lenient, and don't care at all for tags?1 Would it not be better to try and parse with tag verification. if that fails, try again without tag verification and warn, if this succeeds? I mean: this is transitional, is it not? or do we want to permanently remain lenient?

I would go for `be conservative in what you send, be liberal in what you accept`, and not try to parse it twice.

JesusMcCloud (approved):
Your call

PR #143, opened 2024-10-07
Extend RequestParsing to SignatureRequests

n0900:
`SignatureRequest` can be passed by reference just like `AuthenticationRequest`. 
Here we extend `AuthenticationRequestParser` to also handle requests by reference for signatures sent by the driving application/terminal service. For consistency we rename the class to `RequestParser` and introduce the `SignatureRequestFrom` data classes.


nodh:
It's hard to review this, since that other one #127 is still a draft...
But to me the `SignatureRequestParameters` look a lot like `AuthenticationRequestParameters`, so why split it up? Since the latter one still has `numSignatures` and other stuff in it?!

n0900:
> It's hard to review this, since that other one #127 is still a draft... 

Admittedly the draft became quite unwieldy but it is not yet at a point where merging makes sense so I planned on splitting new changes into seperate merge requests and the draft in the end is the full feature implementation. What do you think of this?

> But to me the `SignatureRequestParameters` look a lot like `AuthenticationRequestParameters`, so why split it up? Since the latter one still has `numSignatures` and other stuff in it?!

Since they do not have a lot of fields in common I think it makes sense to keep the two seperate. We also talked about seperating some other classes at a previous time and I thought it would be good to not add more to these. So I introduced the interface `RequestParameter`

We could realistically combine `AuthenticationRequestParameterFrom` and `SignatureRequestParameterFrom` to simply be `RequestparameterFrom` however this has a long rat tail in `OidcSiopWallet` and beyond which was rather unpleasant to work with so I decided against it. It you want I could also implement a version like this.

n0900:
> But to me the `SignatureRequestParameters` look a lot like `AuthenticationRequestParameters`, so why split it up? Since the latter one still has `numSignatures` and other stuff in it?!

On second look I agree that they look related and I have moved the vals they have in common to the interface. However CSC actually extends `AuthenticationRequestParameters` as well and as such uses both in different contexts. This in turn means that combining them is very messy from a readability perspective.

dif-data-classes/src/commonMain/kotlin/at/asitplus/dif/rqes/CollectionEntries/Document.kt
```diff
@@ -23,6 +23,7 @@ import kotlinx.serialization.json.JsonObject
 @Serializable
 data class Document(
     /**
+     * TODO
```
nodh, line 26: still relevant?

vck-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidc/OidcSiopInteropTest.kt
```diff
@@ -246,7 +246,7 @@ class OidcSiopInteropTest : FreeSpec({
         val parsed = jsonSerializer.decodeFromString<AuthenticationRequestParameters>(input)
         parsed.shouldNotBeNull()
 
-        parsed.responseUrl shouldBe "https://verifier-backend.eudiw.dev/wallet/direct_post"
+        parsed.responseUri shouldBe "https://verifier-backend.eudiw.dev/wallet/direct_post"
```
nodh, line 249: Please rename `responseUri` to `responseUrl`, so that we don't introduce a breaking change

nodh:
> > But to me the `SignatureRequestParameters` look a lot like `AuthenticationRequestParameters`, so why split it up? Since the latter one still has `numSignatures` and other stuff in it?!
> 
> On second look I agree that they look related and I have moved the vals they have in common to the interface. However CSC actually extends `AuthenticationRequestParameters` as well and as such uses both in different contexts. This in turn means that combining them is very messy from a readability perspective.

Yeah, now I get it ... it is really not a good idea to combine them

nodh:
It looks to me that the parsed `SignatureRequestParametersFrom` are never used anywhere?

n0900:
> It looks to me that the parsed `SignatureRequestParametersFrom` are never used anywhere?

This is true. It is only used as a container for `SignatureRequestParameters` in order to use `RequestParser` as originally written - the 'From' part is currently ignored. The class will be used when parsing requests which are sent by a [driving application](https://apps.egiz.gv.at/drivingapp/). Tests coming up

nodh (approved):

Issue #20, opened 2023-09-04
Selective Disclosure in `kotlinx.serialization`


iaik-jheher:
For any actual real-life wallet use case, we will need some kind of selective disclosure of credentials.

It would be very nice to extend `kotlinx.serialization` with redactable fields; essentially serialization for a `Redactable<T>` object (essentially a fancy optional) with serialization to JWP or some other (non-binary) merkle tree-like structure.

Not a very well-formed idea yet, but it'd be very convenient to be able to do

```kotlin
@Serializable
data class Credential(
  val firstname: Redactable<String>,
  val lastname: Redactable<String>,
  val dob: Redactable<String>)
```
and then have this serialize to a (partially-redacted) Merkle tree. I am also not sure how to best handle revealing of individual nodes after the fact; this sounds like it might need code generation (akin to what kotlinx.serialization already does?) to be done efficiently.

It's... an idea, not very well-structured yet. @JesusMcCloud told me to post it here either way.

JesusMcCloud:
@iaik-jheher don't we have something ready to use for that?

nodh:
Any news on that?

iaik-jheher:
tech demo exists in the internal gitlab (as `jakob.heher/sd-jwt`), I think it was complete when I left it... it's not the same approach as originally outlined in this issue though

JesusMcCloud:
do we still want to pursue this?

nodh:
I think it's not necessary in VC-K at the moment.

JesusMcCloud:
It would be awesome to open-source what we have, although, if we're not using it, it would be effectively unmaintained

PR #155, opened 2024-10-28
SD-JWT: Parse complex structures

nodh:
Part 2 of a series of MR to handle complex SD-JWT, as discussed in https://github.com/a-sit-plus/vck/issues/20, but without `kotlinx.serialization`.

This PR implements decoding and validating complex SD-JWT disclosures.


JesusMcCloud (approved):
looks solid. but why did we not recycle the solution we had internally?

nodh:
> looks solid. but why did we not recycle the solution we had internally?

Internal solution is really nice, but does not apply to our use case here, where we never have a plain JSON object that needs to be converted into a SD object.

nodh:
Has been merged

PR #163, opened 2024-10-30
Typed JWS/JWE

nodh:
What do you think, is this worth it?

See also https://github.com/a-sit-plus/signum/pull/181


JesusMcCloud:
I think it is worth it! I'll review tomorrow

JesusMcCloud (approved):

nodh:
Got merged into `develop`

PR #164, opened 2024-11-12
Device Signed Item Serialization

nodh:
Device signed items are also namespace-dependent in its serialization, just like the issuer signed items.


JesusMcCloud:
does the spec enforce the ordering here, or will we end up using Obor again?

nodh:
It's a map, so we'll know the identifier before reading the value
________________________________

JesusMcCloud (approved):
lgtm

PR #165, opened 2024-11-13
COSE Signed API

nodh:
Uses type parameter from `CoseSigned` to clarify usages of the payload (and reduce the number of manual tagging payloads).

See also https://github.com/a-sit-plus/signum/pull/194

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 Release 5.2.0:
  - SD-JWT: Validate confirmation claims correctly
- - Adapt to changes in `signum`, i.e. the classes `JwsSigned`, `JweDecrypted`, `CoseSigned` are now typed to their payload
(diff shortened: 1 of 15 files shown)


JesusMcCloud (approved):
reflects the changes, so I'm approving even if the Signum-PR is not yet through

n0900:
I could not find a very convenient way to serialize the classes. 
Using this branch in combination with https://github.com/a-sit-plus/signum/pull/194  if you make the generic nullable and extend the case-distinction all tests seem to work except for `Tag24SerializationTest` where the structure of the encoded issued credential differs by a discriminator `D818` and all else being equal.

Attached are the two outputs
[newEncoded.txt](https://github.com/user-attachments/files/17832826/newEncoded.txt)
[previousEncoded.txt](https://github.com/user-attachments/files/17832831/previousEncoded.txt)
The difference is the discriminator in line 27.

Please let me know if we want to pursue this further

JesusMcCloud (approved):
Good Work! As far as I can see, this covers all our cases

PR #170, opened 2024-11-18
Tmp/rebase sigserv

n0900:
Cleaned version of #127 which also renames the last remaining `_Enum` class.
Big thanks to @JesusMcCloud.
Now without commits from @nodh.


nodh (approved):
Great work!

PR #177, opened 2024-12-11
Release/5.2.0

nodh:


conventions-vclib/src/main/resources/vcLibVersions.properties
```diff
@@ -1,4 +1,4 @@
-signum=3.11.0
+signum=3.11.1
 supreme=0.6.0
```
JesusMcCloud, line 2: could also be 0.6.1, even though it does not matter in this project, as the transitive signum dependency is set pulled up anyways

JesusMcCloud (approved):
lgtm

PR #179, opened 2024-12-16
Fix/dependency cleanups

JesusMcCloud:
* removes obsolete KmmResult and needless arrow implemetation dependencies
* declare arrow as api dependency to work around [KT-73858](https://youtrack.jetbrains.com/issue/KT-73858) 

This allows building valera without issues and without redecalring arrow in valera


nodh (approved):
I've changed base to `develop`, now it looks fine

PR #182, opened 2024-12-18
prepare proper removal of arrow

JesusMcCloud:
This removes the arrow dependency for good.

- We still need to update and release all credentials to use this VC-K version once released
- Then we need pull those updated credentials into Valera (or wherever we use them)
- Finally, our Valera build issues will also be resolved for good.


nodh (approved):
Great!

PR #183, opened 2024-12-19
Feature/more csc data classes

n0900:
Adds data classes to parse credential/info and credential/list requests and responses as defined in CSC API v2.0.0.2 which is necessary to build the DTBS/R in the wallet centric model of UC5


nodh (approved):

nodh:
Please add a changelog entry

PR #189, opened 2025-01-08
Tests for ktor clients

nodh:
Adds tests for our ktor-based clients, used in the App Wallet.

Suggesting to squash & merge.


JesusMcCloud (approved):
lgtm, but i can't really judge everything in detail

PR #190, opened 2025-01-09
Update to latest EU PID credential

JesusMcCloud:
Fixed KLIB resolver dependency resolution bug


JesusMcCloud:
let's see if this really does the trick

JesusMcCloud:
AnnotationSpec seems to be JVM-only

nodh (approved):
:heart:

PR #191, opened 2025-01-10
Refactor creation, verification of presentations

nodh:
Refactores several classes to prepare implementation of ISO/IEC 18013-7 Annex B for OpenID4VP mDoc responses (will be a separate PR).


JesusMcCloud (approved):
Full disclosure: I am counting on the tests to catch anything I might have overlooked, as this is quite large and the bulk of it is just a consequence of limited functional changes

JesusMcCloud:
Well, make iOS build again before merging ;-)

PR #192, opened 2025-01-10
ISO 18013-7 Annex B

nodh:
Implements OpenID4VP for mDoc responses, with device authentication and session transcript from ISO/IEC 18013-7 Annex B.


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/helper/PresentationFactory.kt
```diff
@@ -65,6 +88,63 @@ internal class PresentationFactory(
         }
     }
 
+    /**
+     * Performs calculation of the [SessionTranscript] and [DeviceAuthentication],
+     * acc. to ISO/IEC 18013-5:2021 and ISO/IEC 18013-7:2024, if required in [request] (i.e. it will be encrypted)
+     */
+    @Throws(PresentationException::class, CancellationException::class)
+    private suspend fun calcDeviceSignature(
+        responseWillBeEncrypted: Boolean,
```
JesusMcCloud, line 97: rename to `encryptResponse`?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/VerifierAgent.kt
```diff
@@ -53,7 +54,7 @@ class VerifierAgent(
             ?.let { bytes -> Document.deserialize(bytes).getOrNull() }
         if (document != null) {
             val verifiedDocument = runCatching {
-                validator.verifyDocument(document, challenge)
+                validator.verifyDocument(document, { _, _ -> true })
```
JesusMcCloud, line 57: while I have no doubt that this is correct, it *looks* funky. (same goes for line 67).

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/iso/DeviceAuth.kt
```diff
@@ -13,4 +16,13 @@ data class DeviceAuth(
     val deviceSignature: CoseSigned<ByteArray>? = null,
     @SerialName("deviceMac")
     val deviceMac: CoseSigned<ByteArray>? = null, // TODO is COSE_Mac0
-)
\ No newline at end of file
+) {
+
+    fun serialize() = vckCborSerializer.encodeToByteArray(this)
```
JesusMcCloud, line 21: Just as a discussion point: We have so many classes now that have a `serialize()` method and a `deserialize()` on their companion. Especially everything JWT **needs** those methods because the result is not a valid JSON.
This got me thinking: Do we want to introduce an Interface for that?

JesusMcCloud (changes requested):
minor nitpicks and a discussion point

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidc/helper/PresentationFactory.kt
nodh: We're in the step before actually encrypting the response, and this function also does not encrypt it, so I came up with this name -- can we leave it at that?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/VerifierAgent.kt
nodh: Implemented a fallback for this deprecated (and internally never called method) in 693c3570

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/iso/DeviceAuth.kt
nodh: If the interface doesn't introduce more lines of code and complexity, I'm fine withit.
I also think, that we introduced this methods to call them from Swift code, and it's mostly used as a shortcut internally nowadays.

JesusMcCloud: JWT needs these, other formats don't. Since it is kind of a convention, we could formalise it through an interface.
Most obvious issue with this idea: _naming things is hard_!

JesusMcCloud (approved):

nodh:
Merged into `develop`

Issue #153, opened 2024-10-25
Rename `jsonSerializer` or unify


n0900:
Maybe now is a good time to rename the `jsonSerializer` into something more appropriate (or maybe even merge all into one?)

_Originally posted by @nodh in https://github.com/a-sit-plus/vck/pull/147#discussion_r1814980174_

JesusMcCloud:
does #193 fully address this?

PR #193, opened 2025-01-14
Fix wording for OpenID4VP

nodh:
Get rid of the naming `SIOP` and replace with the correct term `OpenID4VP`, also cleaning up the huge classes a bit.

Fixes https://github.com/a-sit-plus/vck/issues/153


JesusMcCloud (approved):
LGTM

nodh:
Merged into `develop`

PR #195, opened 2025-01-15
SD-JWT: Support creating nested structures with dot notation

nodh:
This will come in handy for https://github.com/a-sit-plus/eu-pid-credential/pull/11


JesusMcCloud (approved):
LGTM

PR #211, opened 2025-02-06
Release 5.3.3

nodh:
- ISO: Fix serialization of device authentication bytes
 - ISO: Fix generating `mdocGeneratedNonce` for device authentication acc. to ISO 18013-7
 - OpenID4VP: Support extracting authentication response from JWS inside JWE

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/PresentationFactory.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/PresentationFactory.kt
@@ -121,12 +121,15 @@ internal class PresentationFactory(
         val deviceAuthentication = DeviceAuthentication(
             type = "DeviceAuthentication",
             sessionTranscript = sessionTranscript,
-            docType = if (responseWillBeEncrypted) docType else "",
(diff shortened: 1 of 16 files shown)


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/PresentationFactory.kt
```diff
@@ -119,9 +124,14 @@ internal class PresentationFactory(
             docType = docType,
             namespaces = deviceNameSpaceBytes
         )
-        Napier.d("Device authentication is ${deviceAuthentication.serialize().encodeToString(Base16())}")
+        // TODO Change type to CoseSigned<DeviceAuthentication>
```
JesusMcCloud, line 127: why todo?

nodh: Rather large change ... so it would keep it as it is for now

JesusMcCloud (approved):
i just checked the release-relevant stuff, a I assume, the changes are all approved from previous PRs

PR #217, opened 2025-02-07
fully integrated encryption

JesusMcCloud:
not much left of the cryptoservice. We could use this state to test encryption and key agreement on iOS, even before publishing signum. Maybe we should, because this way we could still catch any mishaps in signum before we merge there.
Opinions, anyone? @md4096?


vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/KeyMaterial.kt
```diff
@@ -90,6 +93,18 @@ class EphemeralKeyWithoutCert(
     override suspend fun getCertificate(): X509Certificate? = null
 }
 
+/**
+ * Generate new key material with a random key, e.g. used in tests
+ */
+class KeyWithoutCert(
```
nodh, line 99: Isn't there something like `EphemeralKeyWithoutCert()` already?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/jws/JwsService.kt
```diff
@@ -193,29 +199,26 @@ class DefaultJwsService(private val cryptoService: CryptoService) : JwsService {
...
         val intermediateKey = concatKdf(
             z,
             enc,
             header.agreementPartyUInfo,
-            header.agreementPartyVInfo,
-            enc.encryptionKeyLength
+            header.agreementPartyVInfo
         )
-        val key = compositeKey(enc, intermediateKey)
+        require( alg == JweAlgorithm.ECDH_ES )
```
nodh, line 209: But we could do RSA OEAP now?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/KeyMaterial.kt
JesusMcCloud: IIRC the inheritance hierarchy was problematic. I'll re-check

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/jws/JwsService.kt
JesusMcCloud: but this thing here would be asymmetric encryption, which signum does not yet support?!

JesusMcCloud: the ephemeralkeywithoutcert cannot take a predefined key. but that one was only used for testing, so it is gone now

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/KeyMaterial.kt
JesusMcCloud: gone

JesusMcCloud:
someone should really test this in the app. Just be sure, to also set the key agreement flag for EC key generation and delete the old holder key!

n0900 (approved):
I like it

JesusMcCloud:
then we just have to wait for @iaik-jheher to find the time to review the signum PR

JesusMcCloud:
Badum-Tsss!

JesusMcCloud:
I did not yet get to testing it in the app, because first, we'll need snapshot releases of all credentials due to breaking changes in signum and I cannot make any guarantees to get to it before easter, but I will try to at least release the credentials with some distinctive version string as snapshots so testing becomes possible

JesusMcCloud:
please re-review before I build a snapshot of this state. Once this is approved, I will rebuild credentials and test it in the app

JesusMcCloud:
(I will need pointers to an app branch that works with the current development branch)

vck/src/iosMain/kotlin/at/asitplus/wallet/lib/agent/DefaultCryptoService.kt
```diff
@@ -1,58 +0,0 @@
...
-
-/**
- * Default implementation of a crypto service for iOS.
- *
- * The primary goal is to provide a minimal implementation so that unit tests in the `commonTest` module run successfully.
- *
- * Beware: It does **not** implement encryption, decryption, key agreement and message digest correctly.
- */
-@Suppress("UNCHECKED_CAST")
-actual open class PlatformCryptoShim {
```
nodh, line 23: Won't this lead to compilation errors on clients upgrading from vc-k 5.5.0 to 5.6.0? Can we leave this class and deprecate it?

nodh (changes requested):
Probably needs rebasing, now that https://github.com/a-sit-plus/vck/pull/281 has been merged?

vck/src/iosMain/kotlin/at/asitplus/wallet/lib/agent/DefaultCryptoService.kt
JesusMcCloud: you'll need to update anyway, because of signum breaking changes

JesusMcCloud:
I've now rebased and (at least to me) this again raises the question if we even still need the cryptoservice. the testing argument does not apply anymore with function decomposition done. And since signum is breaking stuff anyway, why not seize the opportunity and remove all the legacy cruft right away? As in: make it `DeprecationLevel.Error` and once everything is upgraded in Valera and in the credentials, remove it for good even before the next VC-K release?!

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Release 5.7.0:
    - Remove code elements deprecated in `5.6.0`
  - Holder:
    - Replace `keyPair` with `keyMaterial`
+ - Fully integrated crypto functionality based on Signum 3.16.1
```
nodh, line 10: Please document also the necessary changes for `JwsAlgorithm.ES256` to `JwsAlgorithm.Signature.ES256` (probably others?)

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/CryptoService.kt
```diff
@@ -3,13 +3,22 @@
...
 import at.asitplus.KmmResult
+import at.asitplus.catching
 import at.asitplus.signum.indispensable.CryptoPublicKey
 import at.asitplus.signum.indispensable.CryptoSignature
+import at.asitplus.signum.indispensable.KeyAgreementPublicValue
 import at.asitplus.signum.indispensable.SignatureAlgorithm
-import at.asitplus.signum.indispensable.josef.JweEncryption
+import at.asitplus.signum.indispensable.symmetric.AuthCapability
+import at.asitplus.signum.indispensable.symmetric.NonceTrait
+import at.asitplus.signum.indispensable.symmetric.SealedBox
```
nodh, line 13: why are here additional imports needed, but the code has been removed?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/KeyMaterial.kt
```diff
@@ -11,6 +13,7 @@ import at.asitplus.signum.indispensable.toX509SignatureAlgorithm
 import at.asitplus.signum.supreme.asKmmResult
 import at.asitplus.signum.supreme.sign.EphemeralKey
 import at.asitplus.signum.supreme.sign.Signer
+import at.asitplus.signum.supreme.sign.signerFor
```
nodh, line 16: also here: Deleted code but added imports?

nodh (changes requested):

nodh (approved):
:rocket:

PR #231, opened 2025-03-03
Signum 3.14.0

nodh:
@JesusMcCloud  Is there anything more to do, e.g. for Key Agreement?


JesusMcCloud (approved):
I am surprised that it still builds as a composite build. But if it works, it works! I'll take care of the update to the next Signum release, where some slight changes due to convention updates will be required.

Issue #152, opened 2024-10-24
Improve Simple Authorization Service


nodh:
Store the scope values or authorization details associated with that access token, so it can be verified when issuing a credential.

_Originally posted by @JesusMcCloud in https://github.com/a-sit-plus/vck/pull/148#discussion_r1808257641_

PR #232, opened 2025-03-04
OAuth2 improvements

nodh:
Resolves https://github.com/a-sit-plus/vck/issues/152

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialAuthorizationServiceStrategy.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialAuthorizationServiceStrategy.kt
@@ -41,7 +41,7 @@ class CredentialAuthorizationServiceStrategy(
         }
         .joinToString(" ")
 
-    override fun filterAuthorizationDetails(authorizationDetails: Set<AuthorizationDetails>) =
+    override fun filterAuthorizationDetails(authorizationDetails: Collection<AuthorizationDetails>) =
         authorizationDetails
             .filterIsInstance<OpenIdAuthorizationDetails>()
             .mapNotNull {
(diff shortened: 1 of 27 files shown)


JesusMcCloud (approved):
what is the rationale of preferring a collection over a set? is the order important?

nodh:
> what is the rationale of preferring a collection over a set? is the order important?
A collecltion has no order, that would be a list, right? In the API I've chosen `Collection` as the type, to keep it as generic as possible for clients of that API.

JesusMcCloud:
I'm just wondering: semantically, is it legal and/or does it make sense to allow a generic collection? a list preserves order, but allows duplicates, a set has no guaranteed order, but disallows duplicates. I don't think we need sorting or stable orders. but do we need the ability to allow one item to be present multiple times?

nodh:
> I'm just wondering: semantically, is it legal and/or does it make sense to allow a generic collection? a list preserves order, but allows duplicates, a set has no guaranteed order, but disallows duplicates. I don't think we need sorting or stable orders. but do we need the ability to allow one item to be present multiple times?

I still don't know to which `Collection` you are refering to. But the `filterAuthorizationDetails` for example takes in a `Collection` and returns a `Set`.

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialAuthorizationServiceStrategy.kt
```diff
@@ -34,7 +41,7 @@ class CredentialAuthorizationServiceStrategy(
         }
         .joinToString(" ")
 
-    override fun filterAuthorizationDetails(authorizationDetails: Set<AuthorizationDetails>) =
+    override fun filterAuthorizationDetails(authorizationDetails: Collection<AuthorizationDetails>) =
```
JesusMcCloud, line 44: This function here does not check for duplicates

JesusMcCloud (changes requested):

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialAuthorizationServiceStrategy.kt
nodh: But it returns a set!?

JesusMcCloud: 🤦

JesusMcCloud (approved):

PR #233, opened 2025-03-05
OpenID4VCI Key Attestations

nodh:


openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/KeyAttestationRequired.kt
```diff
@@ -0,0 +1,19 @@
...
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class KeyAttestationRequired(
+    /**
+     * OID4VCI: OPTIONAL. Array defining values specified in Appendix D.2 accepted by the Credential Issuer.
+     */
+    @SerialName("key_storage")
+    val keyStorage: Collection<String>? = null,
```
JesusMcCloud, line 12: Why not make these values an enum?

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/KeyAttestationRequired.kt
```diff
@@ -0,0 +1,19 @@
...
+
+    /**
+     * OID4VCI: OPTIONAL. Array defining values specified in Appendix D.2 accepted by the Credential Issuer.
+     */
+    @SerialName("user_authentication")
+    val userAuthentication: Collection<String>? = null,
```
JesusMcCloud, line 18: Why not make these values an enum?

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/KeyAttestationRequired.kt
nodh: Because I assume that those values will change, and because they have no impact as of yet

JesusMcCloud (approved):

PR #239, opened 2025-03-10
OAuth2: Refresh tokens

nodh:
Follows recommendation to use refresh tokens from https://openid.net/specs/openid4vc-high-assurance-interoperability-profile-1_0.html#name-token-endpoint


JesusMcCloud (commented):
preliminary comment: very clean, but I need to re-review when I have some peace and quiet

acrusage-iaik (approved):
I'm not sufficiently versed in OID4VCI

PR #241, opened 2025-03-13
OID4VCI: Refactor token generation

nodh:


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/SimpleAuthorizationService.kt
```diff
@@ -99,6 +97,9 @@ class SimpleAuthorizationService(
     ),
 ) : OAuth2AuthorizationServerAdapter {
 
+    /** Only for local tests. */
+    private val listOfValidatedAccessToken = mutableListOf<ValidatedAccessToken>()
```
JesusMcCloud, line 101: How and where is it used for local tests **only**? I see tokens getting added and this list being used as fallback. smells a bit funky

JesusMcCloud (changes requested):
Convince me that how `listOfValidatedAccessTokens` is only used for local tests without straying into real business logic or separate it

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/SimpleAuthorizationService.kt
nodh: That has been changed just in the next PR: https://github.com/a-sit-plus/vck/pull/242/files#diff-a29402ee40558eb30dbcca06b7237d84406f1271b1df6e1d3767dca18698a656R90

JesusMcCloud (approved):

PR #242, opened 2025-03-13
OID4VCI: Finalize token responsibilities

nodh:


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialIssuer.kt
```diff
@@ -159,7 +160,8 @@ class CredentialIssuer(
      *
      * @param user used to create the credential when the wallet app requests the credential
      */
-    // TODO Migrate this to OAuth2AuthorizationServerAdapter?
+    @Suppress("DEPRECATION")
```
JesusMcCloud, line 163: ?

JesusMcCloud (approved):

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialIssuer.kt
JesusMcCloud: nvm. although i would prefer the annotation right before the statement that causes a warning

PR #243, opened 2025-03-14
OAuth2: Move User Data Provider

nodh:


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialAuthorizationServiceStrategy.kt
```diff
@@ -29,6 +32,8 @@ class CredentialAuthorizationServiceStrategy(
             OpenIdAuthorizationDetails(credentialConfigurationId = it.key)
         }
 
+    @Suppress("DEPRECATION")
```
JesusMcCloud, line 35: ?

JesusMcCloud: ah! it is for a single statement inside the call…

JesusMcCloud (approved):
looks really clean

Issue #251, opened 2025-03-28
Missing API in "TransactionData.QesAuthorization"


tlenz:
## Issue

OpenId4VP defines that the credential_ids that must be part of a transaction_data element has to be an ID from InputDescriptor.

credential_ids: REQUIRED. Array of strings each referencing a Credential requested by the Verifier that can be used to authorize this transaction. In [[DIF.PresentationExchange](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#DIF.PresentationExchange)], the string matches the id field in the Input Descriptor. In the Digital Credentials Query Language, the string matches the id field in the Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for transaction authorization. transaction_data_hashes_alg: OPTIONAL. Array of strings each

Currently, there is no way to set *credential_ids* and *transaction_data_hashes_alg* in https://github.com/a-sit-plus/vck/blob/1c470d2badda6ed85274e3458ead54b3e18c3ec5/rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/TransactionData.kt#L124

## Solution
 Two ways are possible
1. Extend the creator with with additional parameters: https://github.com/a-sit-plus/vck/blob/1c470d2badda6ed85274e3458ead54b3e18c3ec5/rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/TransactionData.kt#L124
2. Remove *private constructor* from https://github.com/a-sit-plus/vck/blob/1c470d2badda6ed85274e3458ead54b3e18c3ec5/rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/TransactionData.kt#L45 to make it as same as defined for *QCertCreationAcceptance*

nodh:
Does DCQL need to be considered here too?

PR #252, opened 2025-03-28
Rrequest options credential

### LINKED ISSUE #250: Missing API option to set "transaction_data" -> "credential_ids"
## Issue
OpenId4VP defines that the *credential_ids* that must be part of a *transaction_data* element has to be an ID from InputDescriptor.

_credential_ids: REQUIRED. Array of strings each referencing a Credential requested by the Verifier that can be used to authorize this transaction. In [[DIF.PresentationExchange](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#DIF.PresentationExchange)], the string matches the id field in the Input Descriptor. In the Digital Credentials Query Language, the string matches the id field in the Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for transaction authorization. transaction_data_hashes_alg: OPTIONAL. Array of strings each_ 

Currently, there is no way to get the Id before generating the full request. However, the Id is needed to set them into  *transaction_data* element.

## Proposed solution: 
Extend *RequestOptionsCredential* https://github.com/a-sit-plus/vck/blob/1c470d2badda6ed85274e3458ead54b3e18c3ec5/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt#L168  with an *Id* parameter to set the Id from outside and only generate a fresh Id in https://github.com/a-sit-plus/vck/blob/1c470d2badda6ed85274e3458ead54b3e18c3ec5/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestOptions.kt#L191 if no *Id* was set.

nodh:
Fixes #250, #251

### DIFF
--- a/rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/TransactionData.kt
+++ b/rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/TransactionData.kt
@@ -109,7 +109,7 @@ sealed class TransactionData {
          * Validation according to D3.1: UC Specification WP3
          */
         init {
-            require(credentialID or signatureQualifier) {"Either credentialID or signatureQualifier must be set"}
+            require(credentialID or signatureQualifier) { "Either credentialID or signatureQualifier must be set" }
         }
 
         companion object {
@@ -121,10 +121,14 @@ sealed class TransactionData {
                 signatureQualifier: SignatureQualifier? = null,
                 credentialId: String? = null,
                 processID: String? = null,
+                credentialIds: Set<String>? = null,
+                transactionDataHashAlgorithms: Set<String>? = null,
             ): KmmResult<TransactionData> = runCatching {
                 QesAuthorization(
                     signatureQualifier = signatureQualifier,
                     credentialID = credentialId,
+                    credentialIds = credentialIds,
+                    transactionDataHashAlgorithms = transactionDataHashAlgorithms,
                     documentDigests = documentDigest,
                     processID = processID,
                 )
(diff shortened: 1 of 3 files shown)


rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/TransactionData.kt
```diff
@@ -121,10 +121,14 @@ sealed class TransactionData {
                 signatureQualifier: SignatureQualifier? = null,
                 credentialId: String? = null,
                 processID: String? = null,
+                credentialIds: Collection<String>? = null,
+                transactionDataHashAlgorithms: Set<String>? = null,
             ): KmmResult<TransactionData> = runCatching {
                 QesAuthorization(
                     signatureQualifier = signatureQualifier,
                     credentialID = credentialId,
+                    credentialIds = setOfNotNull(credentialId) + (credentialIds ?: listOf()),
```
tlenz, line 130: In that case it's not a combination of _credentialID_ and  _credentialIds_, because they have different meanings.

- **credentialID** is the identifier of the signing-key on QTSP side, It's an identifier for the qualified certificate and it's defined in UC5 / EUDIW Signature.
- **credentialIds** are defined in openId4VP and focus on the verifiable credential linked to this transaction_data element.

Maybe we should rename the Class member variables to make that more clear

tlenz (changes requested):
the `QesAuthorization ` has different meanings for _credentialId_ and _credentialIds_

rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/TransactionData.kt
nodh: Fixed that

tlenz (approved):

n0900 (approved):

PR #255, opened 2025-04-01
Remove deprecations

nodh:


JesusMcCloud (approved):
I think removing all this stuff is fine, because everything is evolving (including the spec) so we don't need long-term compatibility yet

PR #264, opened 2025-04-07
Fix erroneous requirement check

### LINKED ISSUE #262: QSig - Required check on optional element
## Issue
The element `status: CertStatus` in `CertificateParameters` is marked as optional.
However, the `setSigningCredential(...) in `RqesOpenId4VpHolder` implements a required check, see:

https://github.com/a-sit-plus/vck/blob/d2035e71a0e95dbbc1f9fbe3e38e913097d9b06d/vck-rqes/src/commonMain/kotlin/at/asitplus/wallet/lib/rqes/RqesOpenId4VpHolder.kt#L75 

and that always fails if the `status`was not set.

## Solution
Either remove the `require` check or only check `CertificateParameters.CertStatus.VALID` if the status is available.

n0900:
Fixes #262 and improves error responses


tlenz (approved):

nodh (approved):
Please *rebase* and merge

PR #265, opened 2025-04-07
OpenID4VP: Set exact content type for authentication responses

nodh:
Fixes issues with e.g. [MATTR Verifier](https://tools.mattrlabs.com/mdoc-online-presentation)


tlenz (approved):
The implementation looks well.

It's valid that _application/x-www-form-urlencoded_ does not a charset parameter in that case, because UTF-8 is default.

PR #269, opened 2025-04-08
Implement required classes for NFC presentation

gp-iaik:


JesusMcCloud:
this might be wrong, because I have yet to take a closer look, but if it makes sense, you can implement it right away:

i don't have an elegant solution, but a hacky one, if we need both variants: Create two classes: a new one one with the tag and keep the old one. then try-catch deserialize one or the other.

Better solution: custom deserializer and inside the `decode` funcytion, check if you are deserializing cbor (instance-of check on the decoder). if you are: manually check if the tag is there and parse it f it is. skip it otherwise.

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/iso/NFCHandover.kt
JesusMcCloud, line 1: Discussion point: Since this has both `serialize` and `deserialize` I think we should add a cbor serializer object to the companion that explicitly sets all flags so it behaves as intended to make the workaround work even if something should change to the base CBOR serializer. We should also not that in the API docs for `serialize` and `deserialize`. what do you guys think?

JesusMcCloud (commented):
That's my kind of ~~hack~~workaround! See the open discussion point.

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/iso/NFCHandover.kt
nodh: That comment should belong to `SessionTranscript`, right?

nodh (approved):
I've applied some refactorings, also to make the test a bit more readable

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/iso/NFCHandover.kt
gp-iaik: Yes, a lot will break. Do we want to gradually add serializers to these classes (and missing bytestring annotations) starting with this one? I don't have a strong opinion either way, but it will just make sense once every class has been fixed.

nodh: Isn't creating new serializer objects rather expensive, so we might not should do it for every data class companion object?

PR #272, opened 2025-04-09
Fix deserialization of presentation definition

nodh:
Otherwise it fails because no polymorphic serializer is registered for `InputDescriptor`.

@n0900 Please have a look at other classes where this might be the case.


n0900 (commented):
Does this mean we eliminate `.serialize()` and `.deserialize()` functions or should we extend them via `serializer: Json` parameter?

PR #273, opened 2025-04-09
Fix JWT type of status list token

nodh:
See https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-08.html#section-5.1-3.1.1:

The following content applies to the JWT Header:

* `typ`: REQUIRED. The JWT type MUST be `statuslist+jwt`.

### DIFF
--- a/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/IssuerAgent.kt
+++ b/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/IssuerAgent.kt
@@ -3,7 +3,6 @@ package at.asitplus.wallet.lib.agent
 import at.asitplus.KmmResult
 import at.asitplus.catching
 import at.asitplus.signum.indispensable.SignatureAlgorithm
-import at.asitplus.signum.indispensable.asn1.KnownOIDs.contentType
 import at.asitplus.signum.indispensable.cosef.CoseHeader
 import at.asitplus.signum.indispensable.cosef.CoseSigned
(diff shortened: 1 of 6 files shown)


acrusage-iaik:
This change is fine, but I'd like to provide more context:

The token status list references [RFC7519](https://www.rfc-editor.org/rfc/rfc7519).
RFC7519 references both [RFC7515](https://www.rfc-editor.org/rfc/rfc7515.html) regarding the definition of header `typ`.
RFC7515 defines `typ` to define a [Media Type](http://www.iana.org/assignments/media-types), referencing [RFC2045](https://www.rfc-editor.org/rfc/rfc2045), with the following context-specific interpretation:

>    To keep messages compact in common situations, it is RECOMMENDED that
   producers omit an "application/" prefix of a media type value in a
   "typ" Header Parameter when no other '/' appears in the media type
   value.  A recipient using the media type value MUST treat it as if
   "application/" were prepended to any "typ" value not containing a
   '/'.

The media types "application/statuslist+jwt" and "statuslist+jwt" should therefore be treated equivalently by a conforming JWT implementation.
Disallowing "application/statuslist+jwt" therefore imples that this jwt MUST NOT be treated like a JWT as defined in RFC7519, contradicting the definition `The Status List Token MUST be encoded as a "JSON Web Token (JWT)" according to [[RFC7519](https://www.ietf.org/archive/id/draft-ietf-oauth-status-list-10.html#RFC7519)].`

acrusage-iaik (approved):

PR #274, opened 2025-04-09
Fix erroneous encoding via contextual serialization

n0900:
Issue:
In bigger structures, when the specific class was known, kotlinx.serialization would pick the generic polymorphic serializer for `InputDescriptor` instead of the specific `InputDescriptorSerializer` which is defined for the interface. In this case the `type` discriminator was added.

Solution:
To address this we force the serializer to use the contextual serializer instead which we set to the appropriate `InputDescriptorSerializer`.
To this end we also add a `BaseInputDescriptorSerializer` which only calls the `DifInputDescriptor` serializer without adding the type discriminator.
We need contextual serialization because `InputDescriptor` as an interface allows for open polymorphism and in our use case specifically `QesInputDescriptor` is not defined in the DIF artifact which means the Serializer cannot be hard-coded into the `PresentationDefintion` class.

Also added some tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 Release 5.6.0:
+- Fix erroneous `InputDescriptor` encoding in `PresentationDefinition` when more specific type 
+was known (i.e. `DidInputDescriptor`/`QesInputDescriptor`) via contexutal serialziation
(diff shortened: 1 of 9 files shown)


nodh (approved):
nice!

PR #278, opened 2025-04-18
Fix/direct post response jwt

n0900:
In order for the response to a `direct_post.jwt` request to be valid the response must contain a JWT we cannot simply be serialize the data class to JsonString.


nodh:
Can we add a test for this?

nodh (changes requested):
Can we add a test for this?

n0900:
Added test case which gave rise to the error i.e. the keys cannot be received from an URL and the signing algorithm is not specified.
The more I look at it the more I question `AuthenticationResponse.requestsSignature()` because `RelyingPartyMetaData.authorizationSignedResponseAlg` may be omitted but is in that case assumed to be RS256 which is not taken into account here.
Anyhow the default behaviour from `buildJarm()` should be to throw if it arrives in this case.

nodh (approved):

tlenz (approved):

PR #279, opened 2025-04-18
Feature/verify transaction data

n0900:
Addresses the following bugs/shortcomings concerning verification of transaction data as in UC5 or OID4VP encountered during cross-border testing:
- Change transaction data and related data elements from set to list in order to impose a strict order (UC5 and OID4VP)
- Change transaction data elements from their class to JsonPrimitive as de-/encoding from and to Json is not perfectly stable (consider normal vs pretty-print, encoded default values or not etc) (UC5 and OID4VP)
- Add `TransactionDataBase64Uri` typealias for JsonPrimitive. We do not store the transaction data as string as otherwise this leads to escaped strings in nested structures.
- Add transaction data verification to `OpenID4VpVerifier.validateAuthnResponse`. If transaction data contains valid OID4VP transaction data instances then we only verify these. If none are present we assume all remaining entries are UC5 compliant and are verified as described in the relevant spec.

Test vectors have been extended to now contain two instead of one transaction data entry to check the order requirement and a modified test vector which uses different encoding for transaction data to make sure we do not reject valid requests based on representation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,11 @@ Release 5.6.0:
    - Deprecate `RqesOpenId4VpVerifier`
    - Change `TransactionData` from sealed class to interface
    - Fix erroneous `TransactionData` encoding in `AuthenticationRequest`
- - OpenID for Verifiable Credential Issuance:
+   - Change transaction data and related data elements from set to list
+   - Change transaction data elements from their class to JsonPrimitive
+   - Add `TransactionDataBase64Uri` typealias for JsonPrimitive
+   - Add transaction data verification to `OpenID4VpVerifier.validateAuthnResponse`
+  - OpenID for Verifiable Credential Issuance:
    - Remove code elements deprecated in 5.5.0
  - OpenID for Verifiable Presentations:
    - In `OpenId4VpVerifier` add constructor parameter `supportedAlgorithms`
--- a/rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/QCertCreationAcceptance.kt
+++ b/rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/QCertCreationAcceptance.kt
@@ -1,11 +1,15 @@
 package at.asitplus.rqes.collection_entries
 
 import at.asitplus.openid.TransactionData
+import at.asitplus.rqes.rdcJsonSerializer
+import at.asitplus.rqes.serializers.DeprecatedBase64URLTransactionDataSerializer
 import at.asitplus.signum.indispensable.asn1.ObjectIdSerializer
 import at.asitplus.signum.indispensable.asn1.ObjectIdentifier
 import at.asitplus.signum.indispensable.io.ByteArrayBase64Serializer
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonPrimitive
 
 /**
  * D3.1: UC Specification WP3:
@@ -77,6 +81,12 @@ data class QCertCreationAcceptance(
 
 ) : TransactionData {
 
+    override fun toBase64UrlString(): JsonPrimitive =
+        rdcJsonSerializer.parseToJsonElement(
+            rdcJsonSerializer.encodeToString(
+                DeprecatedBase64URLTransactionDataSerializer, this
+            )
+        ) as JsonPrimitive
 
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
--- a/rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/QCertCreationAcceptance.kt
+++ b/rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/QCertCreationAcceptance.kt
@@ -81,7 +81,7 @@ data class QCertCreationAcceptance(
 
 ) : TransactionData {
 
-    override fun toBase64UrlString(): JsonPrimitive =
+    override fun toBase64UrlJsonString(): JsonPrimitive =
         rdcJsonSerializer.parseToJsonElement(
             rdcJsonSerializer.encodeToString(
                 DeprecatedBase64URLTransactionDataSerializer, this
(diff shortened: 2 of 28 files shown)


vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Validator.kt
```diff
@@ -250,6 +244,23 @@ class Validator(
             Napier.w("verifyVpSdJwt: Key Binding does not contain correct sd_hash")
             return VerifyPresentationResult.ValidationError("Key Binding does not contain correct sd_hash")
         }
+        transactionData?.let {
```
tlenz, line 247: Can we add a property into `Validator` that disables this check?
The reason is that many Wallets in Potential do not support `transaction_data` yet, and that will break with any other Wallet implementation.

vck-rqes/src/commonTest/kotlin/at/asitplus/wallet/lib/rqes/KeyBindingTests.kt
```diff
@@ -0,0 +1,171 @@
...
+                    .getOrThrow()
+            ).getOrThrow().toStoreCredentialInput()
+        )
+
+        holderOid4vp = OpenId4VpHolder(
+            holder = holderAgent,
+        )
+
+    }
+
```
tlenz, line 53: There, we only test success cases for `transaction_data`.
We should add a test for error cases, too.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ was known (i.e. `DidInputDescriptor`/`QesInputDescriptor`) via contexutal serial
 - Deprecate `RqesOpenId4VpVerifier`
 - Change `TransactionData` from sealed class to interface
 - Fix erroneous `TransactionData` encoding in `AuthenticationRequest`
+- Change transaction data and related data elements from set to list
+- Change transaction data elements from their class to JsonPrimitive
+- Add `TransactionDataBase64Uri` typealias for JsonPrimitive
+- Add transaction data verification to `OpenID4VpVerifier.validateAuthnResponse`
```
nodh, line 12: Please rebase this branch and fill out changelog again (contains categorization now)

rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/QCertCreationAcceptance.kt
```diff
@@ -77,6 +81,12 @@ data class QCertCreationAcceptance(
 
 ) : TransactionData {
 
+    override fun toBase64UrlString(): JsonPrimitive =
```
nodh, line 84: The function is named `to...String()` but produces a `JsonPrimitive`, that's not ideal

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Validator.kt
nodh: Would suggest a constructor parameter named `skipTransactionDataValidation: Boolean = false`

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/data/Base64URLTransactionDataSerializer.kt
```diff
@@ -13,6 +14,10 @@ import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
 import kotlinx.serialization.json.JsonElement
 
+@Suppress("Unused")
+fun TransactionDataBase64Url.toDataclass() =
```
nodh, line 18: Please rename to `toTransactionData()`

rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/QCertCreationAcceptance.kt
```diff
@@ -77,6 +81,12 @@ data class QCertCreationAcceptance(
...
+        rdcJsonSerializer.parseToJsonElement(
+            rdcJsonSerializer.encodeToString(
+                DeprecatedBase64URLTransactionDataSerializer, this
```
nodh, line 87: Why use the deprecated serializer here, in new code?

nodh (changes requested):

rqes-data-classes/src/commonMain/kotlin/at/asitplus/rqes/collection_entries/QCertCreationAcceptance.kt
n0900: we will need to use it everywhere applicable until we fully remove it otherwise backwards compatability breaks.

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Validator.kt
n0900: I chose `verifyTransactionData: Boolean = true` because it is more in line with other parameter names but can change it if desired.

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/AuthenticationResponseFactory.kt
```diff
@@ -2,19 +2,24 @@ package at.asitplus.wallet.lib.openid
 
 import at.asitplus.openid.*
 import at.asitplus.openid.OpenIdConstants.ResponseMode.*
+import at.asitplus.openid.RelyingPartyMetadata
```
nodh, line 5: nitpick: Adding imports in a file with no new code should not be necessary

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/VerifiablePresentationFactory.kt
```diff
@@ -6,6 +6,7 @@ import at.asitplus.jsonpath.core.NormalizedJsonPath
 import at.asitplus.jsonpath.core.NormalizedJsonPathSegment
 import at.asitplus.openid.dcql.DCQLClaimsQueryResult
 import at.asitplus.openid.dcql.DCQLCredentialQueryMatchingResult
+import at.asitplus.openid.sha256
```
nodh, line 9: nitpick: adding imports in a file with no new code should not be necessary

nodh (approved):
Awaiting review from @tlenz

tlenz (approved):

PR #280, opened 2025-04-22
Add classes for DC API as defined in OID4VP and ISO 18013-7

gp-iaik:


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/AuthorizationRequestValidator.kt
```diff
@@ -27,6 +27,9 @@ internal class AuthorizationRequestValidator(
             throw InvalidRequest("response_type is null")
         }
 
+        if (request is RequestParametersFrom.JwsSigned) {
+            request.parameters.verifyClientIdPresent()
+        }
```
gp-iaik, line 32: @nodh What would be the best way to ignore a set client_id in an unsigned request? Or should we throw an invalid request exception in that case? I'm not sure how to interpret the spec in that case

```
The client_id parameter MUST be omitted in unsigned requests defined in [Appendix A.3.1](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#unsigned_request). The Wallet MUST ignore any client_id parameter that is present in an unsigned request.
```

nodh: I would throw an exception if the `client_id` is set, yes. Maybe we'll also need to extend `RequestParametersFrom` with a case for DCAPI?

PR #284, opened 2025-04-25
Fix: DCQL Query Serialization

acrusage-iaik:


openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/AuthenticationRequestParameters.kt
```diff
@@ -188,7 +187,6 @@ data class AuthenticationRequestParameters(
      * OID4VP: dcql_query: A string containing a JSON-encoded DCQL query as defined in Section 6.
      */
     @SerialName("dcql_query")
-    @Serializable(with = DCQLQueryStringTransformingSerializer::class)
```
nodh, line 191: But that class referenced here is now unused, and we can't parse the "old" (incorrect) way anymore, right?

nodh (changes requested):

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/AuthenticationRequestParameters.kt
acrusage-iaik: Do we want this backwards compatibility?

nodh: Yes, please

PR #288, opened 2025-04-29
DCQL: Quick fix for new claim query format

nodh:
Just a quick fix to parse requests from https://funke.animo.id/

### DIFF
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/dcql/DCQLIsoMdocClaimsQuery.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/dcql/DCQLIsoMdocClaimsQuery.kt
@@ -2,10 +2,15 @@ package at.asitplus.openid.dcql
 
 import at.asitplus.KmmResult
 import at.asitplus.catching
+import at.asitplus.jsonpath.core.NormalizedJsonPathSegment
 import at.asitplus.openid.CredentialFormatEnum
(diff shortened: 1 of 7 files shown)


openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/dcql/DCQLIsoMdocClaimsQuery.kt
```diff
@@ -14,14 +17,16 @@ data class DCQLIsoMdocClaimsQuery(
...
+    override val path: DCQLClaimsPathPointer? = null,
 
     /**
      * OID4VP draft 23: namespace: REQUIRED if the Credential Format is based on the mdoc format
      * defined in [ISO.18013-5]; MUST NOT be present otherwise. The value MUST be a string that
      * specifies the namespace of the data element within the mdoc, e.g., org.iso.18013.5.1.
      */
     @SerialName(SerialNames.NAMESPACE)
-    val namespace: String,
+    val namespace: String? = null,
```
acrusage-iaik, line 29: This seems to be deprecated now, but maybe add as property for convenience?
If this should be preserved for backwards compatibility, please validate compatibility if both namespace and path are specified.

acrusage-iaik (changes requested):
Please also fix tests (and feel free to assign to me)

acrusage-iaik (approved):
looks good!

PR #291, opened 2025-05-06
Refactor initialization of Validator

nodh:
Refactors away the secondary constructors


acrusage-iaik (approved):
Also looking forward to using a proper dependency injection framework

nodh:
imho dependency injection should be a job of the client application using this library

PR #296, opened 2025-05-08
Add: Integration of timeliness validation separation

acrusage-iaik:

PR #298, opened 2025-05-08
Timescope

nodh:
What do you think, would this `TimeScope` increase the readability of the time checks, e.g. when and how to add or substract the time leeway


acrusage-iaik:
Nice! Added idea to https://github.com/a-sit-plus/vck/pull/296

acrusage-iaik:
Nice! Added idea to https://github.com/a-sit-plus/vck/pull/296 , since I think rebasing would just be a hassle

PR #303, opened 2025-05-19
Nested Path Queries

nodh:


nodh:
See also https://github.com/a-sit-plus/vck/issues/304 for future work

md4096 (approved):
lgtm

PR #308, opened 2025-05-27
LocalDateOrInstant

nodh:
Might be used by credential implementations, e.g. EU PID


JesusMcCloud:
Thanks! I'll look into it

nodh:
Other idea: What if `vck` deserializes only some bytes (from `obor`?) and then let the credential libraries decide on how to interpret those bytes?

nodh:
> I was thinking obor too, since it is already in VC-K and used to success in similar situations. is this pressing?

But then we'll need the `LocalDateOrInstant` anyway, for `expiry_date` and `issuance_date`, see [PID rulebook](https://github.com/eu-digital-identity-wallet/eudi-doc-architecture-and-reference-framework/blob/main/docs/annexes/annex-3/annex-3.01-pid-rulebook.md#24-mandatory-metadata-specified-in-cir-20242977): 

> Date (and if possible time) ...

JesusMcCloud:
Wait?! everything's green already?

nodh:
If you're happy with my `length > 10` check and the custom serializer for it, we can leave it at that.

See also #311

JesusMcCloud (approved):
Given that length and format of both types are distinct and there is only ever one valid length for a date which is always shorter than the shortest possible length für an instant, I don't see any issues here.  
I just don't understand how tagging is relevant here at the serializer level as neither requires a specific tag, IIRC.

JesusMcCloud (approved):
LGTM

PR #310, opened 2025-06-02
fix(openId4VP): change input data for calculating transaction_data_ha…

tlenz:
There was a discussion in openID4VP about how _transaction_data_hashes_ have to be calculated. 
Agreement to use text from v23, see https://github.com/openid/OpenID4VP/pull/621


nodh (approved):

n0900 (approved):
Tragic

PR #312, opened 2025-06-03
Add workflow cache

md4096:
- Speedup workflows 🚀

### DIFF
--- a/.github/workflows/test-jvm.yml
+++ b/.github/workflows/test-jvm.yml
@@ -12,6 +12,15 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '17'
+      - name: Cache gradle
+        uses: actions/cache@v4
+        env:
+          cache-name: cache-gradle
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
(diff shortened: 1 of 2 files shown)


.github/workflows/test-jvm.yml
```diff
@@ -12,6 +12,15 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '17'
+      - name: Cache gradle
+        uses: actions/cache@v4
+        env:
+          cache-name: cache-gradle
+        with:
+          path: |
```
nodh, line 20: Why not also `~/.m2/repository`?

.github/workflows/test-jvm.yml
```diff
@@ -12,6 +12,15 @@ jobs:
...
+            ~/.gradle/caches
+            ~/.gradle/wrapper
```
nodh, line 22: If we're caching the wrapper, we'll need to add `gradle/wrapper` (in this repository) to the hash input too, right? So it works correctly when we're upgrade the gradle wrapper version?!

nodh (changes requested):

.github/workflows/test-jvm.yml
md4096: Thats a good point! I will add that

nodh (approved):

PR #333, opened 2025-06-29
update to new conventions

JesusMcCloud:
* Kotlin 2.1.21
* BC 1.81
* KmmResult 1.9.3 by default
* New publishing endpoint (will require updated secrets)


JesusMcCloud:
now everything points to libs.versions.toml

JesusMcCloud:
I updated all publishing secrets (and if we're lucky, I did not even mix up a single one)

PR #338, opened 2025-07-03
fix-typos

thecyberfred:


iaik-jheher:
@JesusMcCloud why is this not automatically re-targeted to development

PR #361, opened 2025-07-30
Revert moving issuing fun from constructor

nodh:
Otherwise it would be too easy to mix up key material for metadata and the key material used to actually sign the credentials


JesusMcCloud (approved):
sounds reasonable

PR #372, opened 2025-08-26
 Rework split between OID4VCI and OAuth 2.0 pt1

nodh:
Improves our OAuth2.0 implementation with token exchange.

This allows us to build a RemoteOAuth2AuthorizationServerAdapter for the credential issuer to use an external AS (that is not necessarily our own implementation).

(Tests may fail here, this is just to split up PR #371)


JesusMcCloud (commented):
I really think we should also go for something like https://github.com/a-sit-plus/signum/issues/335 in VC-K for data classes like these.

PR #373, opened 2025-08-26
WIP: JAR request parameters

n0900:
Currently we use AuthenticationRequestParameters to represent both JAR requests which contain RequestObjects and the AuthenticationRequestParameters which are the RequestObjects. Meaning we use the same class twice, nested, for different purposes. To remedy this I propose `JarRequestParameters` which replaces the JAR parameters in `AuthenticationRequestParameters` and represents any request parameter that is wrapped as JAR. 

`OAuth2Client` now explicitly works with either `AuthenticationRequestParameters` or `JarRequestParameters`.
The `RequestParser` now knows when it works with a JAR request and can resolve it directly (no more need for recursion) however this leads to earlier JWS validation which currently kills the tests. 

What do you think?

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/AuthorizationService.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/AuthorizationService.kt
@@ -4,6 +4,7 @@ import at.asitplus.KmmResult
 import at.asitplus.openid.AuthenticationRequestParameters
 import at.asitplus.openid.JarRequestParameters
 import at.asitplus.openid.PushedAuthenticationResponseParameters
+import at.asitplus.openid.RequestParameters
 import at.asitplus.openid.TokenRequestParameters
 import at.asitplus.openid.TokenResponseParameters
 import at.asitplus.wallet.lib.oidvci.OAuth2Exception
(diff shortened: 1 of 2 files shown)


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/AuthorizationService.kt
```diff
@@ -28,6 +29,24 @@ interface AuthorizationService {
...
+     * Clients send their authorization request as HTTP `POST` with `application/x-www-form-urlencoded` to the AS.
+     *
+     * Responses have to be sent with HTTP status code `201`.
+     *
+     * @param request as sent from the client as `POST`
+     * @param clientAttestation value of the header `OAuth-Client-Attestation`
+     * @param clientAttestationPop value of the header `OAuth-Client-Attestation-PoP`
+     * @return [KmmResult] may contain a [OAuth2Exception]
+     */
+    suspend fun par(
```
nodh, line 44: That's the situation I wanted to avoid, because the same method also exists with `AuthenticationRequestParameters` below

nodh (changes requested):

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/AuthorizationService.kt
n0900: We can merge the two and have a switch statement in the function

PR #381, opened 2025-09-10
DPoP Nonces

nodh:
See https://datatracker.ietf.org/doc/html/rfc9449#name-authorization-server-provid and https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-17.html#section-7.2-4


vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/Extensions.kt
```diff
@@ -1,12 +1,60 @@
...
+
+/** Intermediate class to perform error handling on ktor responses, see [onFailure] and [onSuccess]. */
+sealed class IntermediateResult<R> {
+    class Success<R>(val httpResponse: HttpResponse) : IntermediateResult<R>()
+    class Failure<R>(val result: R) : IntermediateResult<R>()
+}
+
+/** Helper method to perform error handling on ktor responses, see [onSuccess]. */
+@OptIn(ExperimentalContracts::class)
+suspend inline fun <reified R> HttpResponse.onFailure(
```
JesusMcCloud, line 30: do we maybe want to add a parameter (with a default catch-all value) to restrict the error rage or the specific error we want to handle? as in:
`onFailure(400) {…}` or `onFailure(400..500/*IntRange*/) {…}`. Of course, we'd then need a third "branch" to handle all other errors, so what is now `onFailure` as an extension on HttpResponse would also need to be an extension on som other intermediate object in addition to handling all errors from HttpResponse directly. Same goes for `onSuccess`

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/RemoteOAuth2AuthorizationServerAdapter.kt
```diff
@@ -127,27 +139,53 @@ class RemoteOAuth2AuthorizationServerAdapter(
...
+        oauth2Client.applyAuthnForToken(
+            oauthMetadata = oauthMetadata,
+            popAudience = publicContext,
+            resourceUrl = url,
+            httpMethod = HttpMethod.Post,
+            useDpop = true,
+            dpopNonce = dpopNonce
+        )()
+    }.onFailure { response ->
+        dpopNonce(response)?.takeIf { retryCount == 0 }?.let { dpopNonce ->
```
JesusMcCloud, line 177: same here

JesusMcCloud (approved):

PR #400, opened 2025-09-18
OpenID4VP 1.0 DCAPI Parsing

nodh:
Handle parsing DCAPI requests just like the other types

### DIFF
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/RequestParametersFrom.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/RequestParametersFrom.kt
@@ -20,13 +20,16 @@ sealed class RequestParametersFrom<S : RequestParameters> {
         override val parameters: T,
         @SerialName(SerialNames.VERIFIED)
         val verified: Boolean,
+        @SerialName(SerialNames.PARENT)
+        val parent: Url?,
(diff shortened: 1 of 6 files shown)


n0900 (approved):
Very nice

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/RequestParametersFrom.kt
```diff
@@ -36,16 +73,36 @@ sealed class RequestParametersFrom<S : RequestParameters> {
...
+            return "Json(parent='$parent', jsonString='$jsonString', parameters=$parameters)"
         }
     }
+
+    object SerialNames {
+        const val TYPE_JWS_SIGNED = "JwsSigned"
+        const val TYPE_JSON = "Json"
+        const val TYPE_DCAPI_UNSIGNED = "DcApiUnsigned"
+        const val TYPE_DCAPI_SIGNED = "DcApiSigned"
+        const val TYPE_URI = "Uri"
```
JesusMcCloud, line 95: No need to change that if you feel it is overkill, but you could add another layer of nesting to the constants so that type constants and property constants are separated (`TYPE_JWS_SIGNED` and `JWS_SIGNED` in the same object has the most footgun potential)

JesusMcCloud (approved):
LGTM

PR #397, opened 2025-09-17
OpenID4VP 1.0 Session Transcript

nodh:
Use handover and session transcript definitions from [OpenID4VP 1.0](https://openid.net/specs/openid-4-verifiable-presentations-1_0-final.html#name-handover-and-sessiontranscr)

PR #401, opened 2025-09-29
OpenID4VP 1.0: Verify ISO session transcript in old format

nodh:
Makes sure the changes from https://github.com/a-sit-plus/vck/pull/397 are backwards compatible on the verifier side, which should also cover problems from https://github.com/a-sit-plus/valera/pull/391


md4096 (approved):
lgtm

JesusMcCloud (changes requested):
I am assuming the test vectors are kosher. make internals `internal`

Issue #388, opened 2025-09-15
Revisit Hashing Algorithms for Transaction Data in SD-JWT


nodh:
What digests shall be supported for SD-JWT and their KB-JWTs? Currently it is only SHA-256.

_Originally posted by @JesusMcCloud in https://github.com/a-sit-plus/vck/pull/384#discussion_r2348381508_

n0900:
This can be closed right?

PR #403, opened 2025-10-03
SD-JWT: Digest algorithms in KB-JWT and transaction data

n0900:
Implements #388


nodh:
Shouldn't this be covered by https://github.com/a-sit-plus/vck/pull/396 or at least be built upon that?

nodh (approved):

PR #407, opened 2025-10-08
SD-JWT: Digest algorithms

n0900:
Implements `_sd_alg` parameter as defined in https://datatracker.ietf.org/doc/draft-ietf-oauth-selective-disclosure-jwt/  Ch 4.1.1 which allows issuer to specify which digest they want to use for selective disclosure and verification thereof. 
Currently we only support SHA256 and ignore this parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 Release 5.9.0 (unreleased):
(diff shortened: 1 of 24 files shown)


nodh (approved):
I've simplified the code a bit, since I see no harm in always specifying the digest algorithm in SD-JWT we are creating.

PR #410, opened 2025-10-22
Build: Support Java 21

nodh:


JesusMcCloud (approved):
Looks good! Did I mess up the ZLibService?

nodh:
> Looks good! Did I mess up the ZLibService?

No, but it doesn't work for JVM (in the old way) if there is no Android SDK installed

PR #426, opened 2025-11-18
OID4VCI: Handle credential response encryption

nodh:
See https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-encrypted-credential-reques

### DIFF
--- a/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/jws/JwsService.kt
+++ b/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/jws/JwsService.kt
@@ -181,7 +181,7 @@ class EncryptJwe(
         val jweEncryption = header.encryption
             ?: throw IllegalArgumentException("No encryption in JWE header")
         val jweHeader = header.copy(
-            jsonWebKey = keyMaterial.jsonWebKey,
+            jsonWebKey = recipientKey,
             ephemeralKeyPair = ephemeralKeyPair.publicValue.asCryptoPublicKey().toJsonWebKey()
         )
         JweUtils.encryptJwe(ephemeralKeyPair, recipientKey, jweEncryption, jweHeader, payload)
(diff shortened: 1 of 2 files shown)


vck-openid-ktor/src/commonTest/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClientTest.kt
```diff
@@ -255,17 +256,25 @@ val OpenId4VciClientTest by testSuite {
                     val requestBody = request.body.toByteArray().decodeToString()
                     val authn = request.headers[HttpHeaders.Authorization].shouldNotBeNull()
                     val params = vckJsonSerializer.decodeFromString<CredentialRequestParameters>(requestBody)
+                    // TODO Test with encryption
```
JesusMcCloud: TODO?

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/WalletService.kt
```diff
@@ -251,15 +259,43 @@ class WalletService(
...
         scheme: ConstantIndex.CredentialScheme,
     ): KmmResult<Collection<Holder.StoreCredentialInput>> = catching {
-        response.extractCredentials()
+        response.decryptResponse()
+            .extractCredentials()
             .map { encryptionService.decrypt(it).getOrThrow() }
             .map { it.toStoreCredentialInput(representation, scheme) }
     }
 
+    private suspend fun CredentialResponse.decryptResponse() = when (this) {
```
JesusMcCloud: Why's it called `decryptResponse` then the Receiver is a response? I'd suggest `decrypt`, ` decryptIfNeeded`, or something like that

vck-openid-ktor/src/commonTest/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClientTest.kt
nodh: See https://github.com/a-sit-plus/vck/pull/434

PR #431, opened 2025-11-19
OID4VP: Evaluate supported VP formats of verifier

nodh:
See https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-metadata and https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-metadata-2 and https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-metadata-3

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/PresentationFactory.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/PresentationFactory.kt
@@ -4,9 +4,7 @@ import at.asitplus.KmmResult
 import at.asitplus.catching
 import at.asitplus.dcapi.DCAPIHandover
 import at.asitplus.dcapi.OpenID4VPDCAPIHandoverInfo
-import at.asitplus.dcapi.request.Oid4vpDCAPIRequest
 import at.asitplus.dif.ClaimFormat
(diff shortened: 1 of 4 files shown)


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/PresentationFactory.kt
```diff
@@ -308,30 +281,45 @@ internal class PresentationFactory(
...
+    ClaimFormat.MSO_MDOC ->
+        if (msoMdoc != null) {
+            var result = true // empty object is fine
+            if (msoMdoc!!.issuerAuthAlgorithms != null) {
+                result = result and msoMdoc!!.issuerAuthAlgorithms!!.any { supportedCoseAlgorithms.contains(it) }
+            }
+            if (msoMdoc!!.deviceAuthAlgorithms != null) {
+                result = result and msoMdoc!!.deviceAuthAlgorithms!!.any { supportedCoseAlgorithms.contains(it) }
+            }
+            result
```
JesusMcCloud: can't you use let + elvis to get rid of circumventing null checks?

JesusMcCloud (changes requested):
why is this making properties optional? shouldn't they be there?

JesusMcCloud (approved):

PR #432, opened 2025-11-20
OID4VP: Correct Wallet metadata

nodh:


openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/OpenIdConstants.kt
```diff
@@ -292,6 +301,15 @@ object OpenIdConstants {
             private const val STRING_FRAGMENT = "fragment"
             private const val STRING_DC_API = "dc_api"
             private const val STRING_DC_API_JWT = "dc_api.jwt"
+
+            fun entries() = setOf(
```
JesusMcCloud: why is this a fun and not a lazy val?

JesusMcCloud (approved):

PR #435, opened 2025-11-24
OID4VP: Always parse authn response correctly

nodh:
Correctly handles authn responses in OpenID4VP.


nodh:
Commits related to removal of `mdocGeneratedNonce` have been moved to  https://github.com/a-sit-plus/vck/pull/439

PR #450, opened 2025-12-02
Build: Fix exports for XCFrameworks

nodh:


dif-data-classes/build.gradle.kts
```diff
@@ -26,6 +32,7 @@ kotlin {
             dependencies {
                 implementation(project.napier())
                 implementation(project.ktor("http")) {
+                    // TODO is this still needed? and why only in this project?
```
nodh: 👀 @JesusMcCloud

JesusMcCloud: removed it and successfully exported locally

PR #464, opened 2025-12-17
OID4VCI: Advertise encryption keys correctly

### LINKED ISSUE #460: Credential Response Encryption
We currently try to get vck running with our PID issuer (which is [the EUDI kotlin issuer](https://github.com/eu-digital-identity-wallet/eudi-srv-pid-issuer)).

We identified two main issues atm:
1. The backend expects an `alg` parameter in the credential response encryption jwk.
2. As soon as this is present the used `enc` must meet the requirements of the issuer. (In our case `A128GCM` instead of `A256GCM`)

Regarding the test setup: We use our own implementation but mainly try to reproduce the issues with the Valera app.

### Ad. 1

> alg (Algorithm), and other JWK parameters. The alg parameter MUST be present.

According to [the specs](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-10) the alg parameter must be present in the credential response encryption jwk.

We where able to achieve that by overwriting the `jsonWebKey` method in our key material.
```kotlin
override val jsonWebKey: JsonWebKey
        get() = super.jsonWebKey.copy(algorithm = JweAlgorithm.ECDH_ES)
```

### Ad. 2

This can also be adjusted from consuming side, but shouldn't both values be derived from the issuer metadata?

```kotlin
WalletEncryptionService(supportedJweEncryptionAlgorithm = JweEncryption.A128GCM, ...)
```

nodh:
Fixes #460 by advertising encryption keys for OID4VCI correctly and pickinge the correct encryption algorithm.

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/WalletEncryptionService.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/WalletEncryptionService.kt
@@ -40,8 +40,10 @@ class WalletEncryptionService(
     private val encryptCredentialRequest: EncryptJweFun = EncryptJwe(EphemeralKeyWithoutCert()),
     /** Algorithms to indicate support for credential response encryption. */
     private val supportedJweAlgorithm: JweAlgorithm = JweAlgorithm.ECDH_ES,
-    /** Algorithms to indicate support for credential response encryption. */
+    @Deprecated("Use fallbackJweEncryptionAlgorithm instead.")
     private val supportedJweEncryptionAlgorithm: JweEncryption = JweEncryption.A256GCM,
+    /** Algorithm to fallback to for credential response encryption. */
+    private val fallbackJweEncryptionAlgorithm: JweEncryption = JweEncryption.A256GCM,
     /** Key to offer for credential response encryption. */
     private val decryptionKeyMaterial: KeyMaterial = EphemeralKeyWithoutCert(),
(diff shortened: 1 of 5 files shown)


tlenz (approved):

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/WalletEncryptionService.kt
```diff
@@ -81,31 +90,34 @@ class WalletEncryptionService(
...
-        } ?: supportedJweAlgorithm
+        }
 
-    /** Fallback to [supportedJweEncryptionAlgorithm] and let's see if issuer can decrypt it. */
-    private fun SupportedAlgorithmsContainer?.selectEncryption(): JweEncryption =
+    /** Try to find a matching encryption algorithm from issuer's metadata. */
+    private fun SupportedAlgorithmsContainer?.selectEncryption(): JweEncryption? =
         this?.supportedEncryptionAlgorithms?.firstOrNull {
-            it == supportedJweEncryptionAlgorithm
+            it == fallbackJweEncryptionAlgorithm
```
JesusMcCloud, line 102: Is this precedence correct?

nodh: I would say so, because we'll try to use "our" algorithm first, if it is supported.

JesusMcCloud (approved):
Trusting you on the precedence

PR #465, opened 2025-12-17
Token status: Use correct CBOR encoding

nodh:
Fixes https://github.com/a-sit-plus/valera/issues/413 by correctly encoding the payload of a status list in CWT, i.e. without tagging it in COSE (24, or in hex `d818`).


JesusMcCloud (approved):
looking really clean

PR #480, opened 2026-01-16
Add tests for StatusList compression fallback

nodh:
### Motivation
- Ensure `StatusList` and `StatusListView` correctly use a `ZlibService` when available and fall back to raw bytes when compression/decompression returns `null`.
- Validate that byte arrays are preserved according to the expected fallback behavior.
- Verify that `statusBitSize` is preserved across conversions between `StatusList` and `StatusListView`.

### Description
- Add `vck/src/commonTest/kotlin/at/asitplus/wallet/lib/extensions/StatusListCompressionTest.kt` containing unit tests for compression/decompression fallbacks.
- Introduce a minimal `FakeZlibService` implementing `ZlibService` that can return `null` for `compress` to force fallback and return known bytes for `decompress` to verify use.
- Add tests asserting `StatusList.toView` uses `decompress` when available and falls back to raw bytes when it returns `null`.
- Add tests asserting `StatusListView.toStatusList` uses `compress` when available and falls back to raw bytes when it returns `null`, while checking `statusBitSize` preservation.

### Testing
- Added unit tests to the repository under `vck/src/commonTest/kotlin/at/asitplus/wallet/lib/extensions/StatusListCompressionTest.kt`.
- No automated test suite (`gradle test` or similar) was executed as part of this change.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69654440c6ac8330b521dd5bf5244d65)

### DIFF
--- a/vck/src/commonTest/kotlin/at/asitplus/wallet/lib/extensions/StatusListCompressionTest.kt
+++ b/vck/src/commonTest/kotlin/at/asitplus/wallet/lib/extensions/StatusListCompressionTest.kt
@@ -1,0 +1,88 @@
+package at.asitplus.wallet.lib.extensions
+
+import at.asitplus.testballoon.invoke
+import at.asitplus.testballoon.minus
+import at.asitplus.wallet.lib.ZlibService
+import at.asitplus.wallet.lib.data.rfc.tokenStatusList.StatusList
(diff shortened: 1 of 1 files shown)


acrusage-iaik (approved):
OIP-560576445

JesusMcCloud (approved):
lgtm

PR #497, opened 2026-01-28
Simplify status list serialization

n0900:
Removes surrogate serialization and replaces it with single custom serializer.
Caveat is that in the current version of Kotlinx.Serialization it is impossible to write a custom serializer which serializes to `ByteString` and so needs
```    
Cbor {
    alwaysUseByteString = true
}
```
to function correctly. This however is not new and already required by other parts of the code base (also documented in openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/iso18013/Identifier.kt)


JesusMcCloud (approved):
🤯👍👍👍

nodh:
@codex review

vck/src/commonTest/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/StatusListSerializationTest.kt
```diff
@@ -152,12 +151,12 @@ val StatusListSerializationTest by testSuite {
...
             )
         ) { (cborString, expectedStatusList) ->
-            val statusList = Cbor.decodeFromHexString<StatusList>(cborString)
+            val statusList = coseCompliantSerializer.decodeFromHexString<StatusList>(cborString)
             expectedStatusList.forEachIndexed { index, it ->
                 statusList.toView()[index.toULong()] shouldBe it
             }
 
-            Cbor.decodeFromHexString<StatusList>(Cbor.encodeToHexString(statusList)) shouldBe statusList
+            coseCompliantSerializer.decodeFromHexString<StatusList>(coseCompliantSerializer.encodeToHexString(statusList)) shouldBe statusList
```
nodh: shouldn't we also check if `coseCompliantSerialize.encodeToHexString(statusList)` matches the `cborString`?

n0900: Screenshot from 2026-01-29 14-22-17

We can - but with all classes which are serialized we run into the problem of parameter order which produce two equivalent but different strings.
(if added right now the test would fail for exactly this reason)

nodh (approved):

vck/src/commonTest/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/StatusListSerializationTest.kt
nodh: Nevertheless I've extended the test to at least check that the list has been encoded correctly

PR #499, opened 2026-01-29
Release 5.11.0

nodh:


JesusMcCloud (changes requested):
* gradle.properties still uses `-SNAPSHOT`
* README button still mentions Kotlin 2.2.21, when in reality it uses 2.3.0

JesusMcCloud (approved):
Make it so!

PR #500, opened 2026-01-30
Introduce interface for storing referenced tokens for status lists

nodh:


JesusMcCloud (approved):
timePeriod seems a strange name at first glance, but it makes sense

md4096 (approved):
lgtm

PR #507, opened 2026-02-10
Propagate ISO mdoc Document.errors into parsed result

nodh:
### Motivation
- Preserve ISO mdoc `Document.errors` so downstream consumers can inspect document-level error codes instead of rejecting documents that include error maps. 
- Avoid hard failure in `ValidatorMdoc.verifyDocument` when `Document.errors` is present, enabling more flexible handling and diagnostics. 
- Record the behavior change in the 5.11.0 changelog so release notes reflect the new behavior.

### Description
- Add `documentErrors: Map<String, Map<String, Int>> = emptyMap()` to `IsoDocumentParsed` to carry extracted `Document.errors`.
- Stop requiring `document.errors == null` in `ValidatorMdoc.verifyDocument` and instead capture `document.errors.orEmpty()` and pass it into the returned `IsoDocumentParsed`.
- Add a unit test `document errors are preserved in parsed output` in `ValidatorMdocTest` that constructs a `Document` with `errors` and asserts `ValidatorMdoc.verifyDocument` preserves them.
- Update `CHANGELOG.md` under Release `5.11.0` to document the ISO mdoc behavior change.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_697a2e37cfb08330a47e25cf831eb6b5)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 Release 5.12.0 (unreleased):
  - Remove code elements deprecated in 5.11.0
+ - ISO mdoc:
+   - Preserve `Document.errors` in parsed ISO document results instead of failing validation
 
 Release 5.11.0:
  - Digital Credentials API:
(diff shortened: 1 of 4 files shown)


JesusMcCloud (changes requested):
i want a clear indicator (property with a custom getter, or whatever) that tells me whether a parsed document is valid or has errors

JesusMcCloud (approved):

PR #511, opened 2026-02-16
Remove custom native ZLIB implementation

nodh:


JesusMcCloud (commented):
I like the idea, but okio bootstraps a whole universe and there is also kotlinx.io, which is okio but different and incompatible so… not sure

nodh:
> I like the idea, but okio bootstraps a whole universe and there is also kotlinx.io, which is okio but different and incompatible so… not sure

Okio's jar seems to be only [75 KB](https://mvnrepository.com/artifact/com.squareup.okio/okio/3.15.0) large, compared to the [50 KB](https://mvnrepository.com/artifact/org.jetbrains.kotlinx/kotlinx-serialization-core/1.10.0) of serialization-core alone.

JesusMcCloud:
we already have koltinx-io in the classpath, and I don't like two subtly incompatible libraries with classes named the same on the same classpath

JesusMcCloud:
Didn't you manually impelemnt Zlib from scratch four years ago?

nodh:
Closed in favor of https://github.com/a-sit-plus/vck/pull/527

Issue #458, opened 2025-12-09
Iso 18013-5 nullability updates


n0900:
In the new version of iso 18013-5 some data classes seem to have updated nullability properties. 
Needs thorough check of data classes. Idk if a sufficiently detailed changelog exists.
Cmp 12.7.1 `NFCTransscript` and #457

PR #514, opened 2026-02-17
ISO: Update data classes to 18013-5 2026

nodh:
Fixes https://github.com/a-sit-plus/vck/issues/458

Tagging also @opsiv

### DIFF
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/iso/EncryptedDocuments.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/iso/EncryptedDocuments.kt
@@ -1,0 +1,41 @@
+package at.asitplus.iso
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.cbor.ByteString
+
+
+/**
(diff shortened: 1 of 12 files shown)


n0900 (approved):

openid-data-classes/src/commonMain/kotlin/at/asitplus/iso/EncryptedDocuments.kt
```diff
@@ -0,0 +1,40 @@
...
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.cbor.ByteString
+
+
+/**
+ * Part of the ISO/IEC 18013-5:2026 standard: Document response encryption (10.3.5)
+ */
+@OptIn(ExperimentalUnsignedTypes::class)
+@Serializable
+data class EncryptedDocuments(
```
gp-iaik, line 13: Is this different from EncryptedResponseData?

nodh: Seems to be slightly different, and also referenced in `DeviceResponse`, so it would keep/add this class here.

PR #515, opened 2026-02-17
Do not throw when matching credentials

acrusage-iaik:


nodh:
@codex review

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/PresentationDataClasses.kt
```diff
@@ -130,28 +130,11 @@ sealed class CreatePresentationResult {
 }
 
 @Serializable
-data class PresentationExchangeCredentialDisclosure(
-    val credential: SubjectCredentialStore.StoreEntry,
+data class PresentationExchangeCredentialDisclosure<Credential: Any>(
+    val credential: Credential,
```
nodh, line 134: Why isn't this a `StoreEntry` anymore?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
```diff
@@ -94,7 +93,7 @@ interface Holder {
         fallbackFormatHolder: FormatHolder? = null,
         pathAuthorizationValidator: PathAuthorizationValidator? = null,
         filterById: String? = null
-    ): KmmResult<Map<String, InputDescriptorMatches>>
+    ): KmmResult<HolderPresentationExchangeQueryMatchingResult<SubjectCredentialStore.StoreEntry>>
```
nodh, line 96: Can we find a way to deprecate the old method and convert to the new one?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
```diff
@@ -121,6 +120,6 @@ interface Holder {
     suspend fun matchDCQLQueryAgainstCredentialStore(
         dcqlQuery: DCQLQuery,
         filterById: String? = null
-    ): KmmResult<DCQLQueryResult<SubjectCredentialStore.StoreEntry>>
+    ): KmmResult<HolderDCQLQueryMatchingResult<SubjectCredentialStore.StoreEntry>>
```
nodh, line 138: Can we find a way to deprecate the old method and convert to the new one?

nodh (changes requested):

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
acrusage-iaik: We can, but I don't think it's necessary, the class HolderDCQLQueryMatchingResult contains the property `credentialQueryMatches` with exactly the same name and semantics, so the only thing that changes for library users is the return type. In this sense the new code is "backwards compatible"

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/PresentationDataClasses.kt
acrusage-iaik: No need to have a fixed type here, fits better in the templated class hierarchies

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/Holder.kt
acrusage-iaik: Done

nodh (approved):

acrusage-iaik:
May I assume you'll be the one to merge this, as it is now assigned to you?

PR #517, opened 2026-02-18
Fix iOS composite build

gp-iaik:


nodh:
I see that the commit https://github.com/a-sit-plus/mobile-driving-licence-credential/tree/6ff1d11e65470f4685f1bf6661b4a05b5964de22 exists ... so why won't the Test JVM workflow check it out?

CI did come to its senses.

nodh (approved):

PR #519, opened 2026-02-23
Add IdentifierList Revocation Mechanism to Credential Issuance

n0900:
(Would replace #504)

Add logic necessary to established classes to do revocation using identifiers instead of doing a full refactor.

Extends Interfaces and classes such that every implementation always does both `StatusList` and `IdentifierList` simultaniously.
Also adds tests which do not require verification because verification is out of scope for this PR.

Update: Added Verification and further tests


nodh:
@codex review

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/InMemoryIssuerCredentialStore.kt
```diff
@@ -81,21 +87,41 @@ class InMemoryIssuerCredentialStore(
...
         val entry = credentialMap.getOrPut(timePeriod) {
             mutableListOf()
         }.find {
             it.statusListIndex == index
         } ?: return false
 
         entry.status = status
+        if (status == TokenStatus.Invalid) {
+            identifierRevokationList.getOrPut(timePeriod) { mutableSetOf() }.add(entry.vcId)
+        }
```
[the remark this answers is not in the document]
n0900: To quote the spec
```
For the identifier list, it implies that only revoked MSOs (as opposed to for example
temporarily suspended MSOs) are put on the identifier list.
Once an MSO is revoked, the MSO shall be permanently revoked and as long as the MSO is still considered
valid according to the validUntil element, no MSO revocation list shall be issued that indicates a
different status than revoked.
```
This behavior is expected.
Drift from the validity defined by `StatusList` must be handled by the consumer

openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/MediaTypes.kt
```diff
@@ -22,6 +22,8 @@ data object MediaTypes {
         /** `application/statuslist+cbor` */
         const val STATUSLIST_CBOR = MediaTypes.Application.STATUSLIST_CBOR
 
+        const val IDENTIFIERLIST_CWT = MediaTypes.Application.IDENTIFIERLIST_CWT
```
nodh, line 32: please add a comment so that it is clear during auto completion in the IDE what value this constant has (see above)

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/InMemoryIssuerCredentialStore.kt
```diff
@@ -27,7 +30,10 @@ class InMemoryIssuerCredentialStore(
     /** Maps timePeriod to credentials */
     private val credentialMap = mutableMapOf<Int, MutableList<Credential>>()
 
-    override suspend fun createStatusListIndex(
+    /** Tracks revoked identifiers for timePeriod to build [IdentifierList]; Sets to remove duplicates */
+    private val identifierRevokationList = mutableMapOf<Int, MutableSet<String>>()
```
nodh:
```suggestion
    private val identifierRevocationList = mutableMapOf<Int, MutableSet<String>>()
```

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/InMemoryIssuerCredentialStore.kt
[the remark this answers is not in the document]
nodh: But would not hurt to add this info as an in-code documentation

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/IssuerCredentialStore.kt
```diff
@@ -16,10 +14,10 @@ interface IssuerCredentialStore {
     )
 
     /**
-     * Called by an [Issuer] when creating a new credential to get a `statusListIndex` first.
+     * Called by an [Issuer] when creating a new credential to get a `statusListIndex` and `identifier first.
      * [Issuer] will call [updateStoredCredential] with the issued credential afterwards.
      */
-    suspend fun createStatusListIndex(
+    suspend fun createStoredCredentialReference(
```
nodh, line 30: Please keep the old method name as deprecated function delegating to the new one

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/IssuerAgent.kt
```diff
@@ -86,16 +89,23 @@ class IssuerAgent(
...
+        val credentialStatus = when(credential.revocationKind) {
+            RevocationList.Kind.STATUS_LIST -> StatusListInfo(
+                index = reference.statusListIndex,
+                uri = UniformResourceIdentifier(getStatusListUrlFor(timePeriod)),
+            )
+
+            RevocationList.Kind.IDENTIFIER_LIST -> IdentifierListInfo(
+                identifier = reference.id.encodeToByteArray(),
+                uri = UniformResourceIdentifier(getIdentifierListUrlFor(timePeriod)),
+                certificate = null //TODO
```
nodh, line 119: We should probably move that out to a `StatusListIssuer` in a future PR?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/StatusListIssuer.kt
```diff
@@ -25,6 +25,12 @@ interface StatusListIssuer : StatusIssuer, StatusProvider {
      * Sets the status of one specific credential to [at.asitplus.wallet.lib.data.rfc.tokenStatusList.primitives.TokenStatus.Invalid].
      * Returns true if this credential has been revoked.
      */
-    fun revokeCredential(timePeriod: Int, statusListIndex: ULong): Boolean
+    fun revokeCredentialByIndex(timePeriod: Int, statusListIndex: ULong): Boolean
```
nodh, line 36: Please keep the old function name as a deprecated method delegating to the new one.

nodh (changes requested):
Great, just some minor remarks

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/IssuerAgent.kt
n0900: Well its not so easy because according to draft-ietf-oauth-status-list the `StatusListIssuer` is a level above the `Issuer`. Moving it up might cause a cyclical dependency

nodh: Right now `StatusListAgent` has no reference to `Issuer`, and `IssuerAgent` has no reference to `StatusListIssuer`, so I think we can find a way.

nodh: See #529

nodh (approved):

PR #527, opened 2026-02-26
Implement ZLIB in pure Kotlin

nodh:
Implemented by gpt-5.3-codex high


vck/src/commonTest/kotlin/at/asitplus/wallet/lib/DefaultZlibServiceOkioInteropTest.kt
```diff
@@ -0,0 +1,61 @@
...
+import io.kotest.matchers.nulls.shouldNotBeNull
+import io.kotest.matchers.shouldBe
+import okio.Buffer
+import okio.buffer
+import okio.deflate
+import okio.inflate
+import kotlin.random.Random
+
+val DefaultZlibServiceOkioInteropTest by testSuite {
+    "ours compressed data inflates with okio" {
```
JesusMcCloud: i'd wrap all these tests inside a `forAll` with bytearray arbs. this would also cover edge cases (while allowing to exclude zero) by default. otherwise it looks sane and if the tests check out it should be fine

JesusMcCloud (changes requested):

JesusMcCloud (approved):

Issue #311, opened 2025-06-02
CBOR Deserializer fallbacks


nodh:
As uncovered by #308, the fallback in `IssuerSignedItemSerializer` doesn't really work. When `CborCredentialSerializer.decode` fails, we should not call `decodeStringElement`, as this will read the next CBOR element (probably having the value `elementIdentifier`).

nodh:
The problem is, that the fallback doesn't work at all, so it might happen in other cases, where the data is encoded wrong.

JesusMcCloud:
what should happen instead? return a generic CBOR element from the Obor classes?

nodh:
See also https://github.com/Kotlin/kotlinx.serialization/issues/2975

JesusMcCloud:
see also https://github.com/Kotlin/kotlinx.serialization/discussions/3165

PR #531, opened 2026-03-02
Robust CBOR deserialization for IssuerSignedItem/List with deferred element identifier

nodh:
Mitigate errors like #311 by preventing deserialization of `IssuerSignedItem` without a namespace. Our code relies on deserializing `IssuerSignedList` instead, which has two advantages: Always has a `namespace` and uses OBOR generic CBOR parsing `elementValue` which should at least return the CBOR byte array as a fallback.

Started with the help of codex 5.3 high.


openid-data-classes/src/commonMain/kotlin/at/asitplus/iso/IssuerSignedItemSerializer.kt
```diff
@@ -90,55 +99,43 @@ open class IssuerSignedItemSerializer(private val namespace: String, private val
...
-                if (random != null && elementValue != null) break
-            }
-        }
-        return IssuerSignedItem(
-            digestId = digestId,
-            random = random!!,
-            elementIdentifier = elementIdentifier,
-            elementValue = elementValue!!,
-        )
+        throw IllegalArgumentException("Do not use IssuerSignedItemSerializer for deserializing, but IssuerSignedItemListSerializer")
```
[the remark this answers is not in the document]
nodh: Valid concern, but it's less confusing if we prevent direct deserialization at all.

openid-data-classes/src/commonMain/kotlin/at/asitplus/iso/IssuerSignedItemSerializer.kt
```diff
@@ -90,55 +99,44 @@ open class IssuerSignedItemSerializer(private val namespace: String, private val
...
 
-        // These are the ones that map to different CBOR data types, the rest don't, so if it is not registered, we'll
-        // lose type information. No others must be added here, as they could consume data from the underlying bytes
-        catchingUnwrapped { return decodeStringElement(descriptor, index) }
-        catchingUnwrapped { return decodeLongElement(descriptor, index) }
-        catchingUnwrapped { return decodeDoubleElement(descriptor, index) }
-        catchingUnwrapped { return decodeBooleanElement(descriptor, index) }
+        val elementValueContainer = first { (it.key as CborText).value == PROP_ELEMENT_VALUE }.value
+        val elementValue = CborCredentialSerializer.lookupSerializer(namespace, elementId)?.let {
+            coseCompliantSerializer.decodeFromByteArray(it, elementValueContainer.cbor)
```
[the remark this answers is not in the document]
nodh: Fixed in f70bc527b

nodh:
pinging @JesusMcCloud

openid-data-classes/src/commonMain/kotlin/at/asitplus/iso/IssuerSignedItemSerializer.kt
```diff
@@ -92,53 +104,108 @@ open class IssuerSignedItemSerializer(private val namespace: String, private val
...
+        }
+
+        throw IllegalArgumentException("Could not decode value at $index for $elementIdentifier")
     }
+
+    internal fun deserializeFromOborMap(item: CborMap): IssuerSignedItem = item.toIssuerSignedItem()
+
+    private fun CborMap.toIssuerSignedItem(): IssuerSignedItem {
+        val digestId = coseCompliantSerializer.decodeFromByteArray(
+            Long.serializer(),
```
JesusMcCloud: long or ulong? important for KxS 1.10

JesusMcCloud (approved):
Looking good.

openid-data-classes/src/commonMain/kotlin/at/asitplus/iso/IssuerSignedItemSerializer.kt
nodh: In the CDDL it's `uint`, so major type 0. But there is no `encodeUIntElement`, so is it safe to use `encodeLongElement` or not?

JesusMcCloud: Isn't there an `ULong.serializer()`?

nodh: But `digestId` is an `UInt`, is it really safe to deserialize it as `ULong`?

JesusMcCloud: we need to wait for the next KxS version anyhow to have number parsing fixed. I'll then look into this version of the CBOR decoder in KxS. but it should probably be ULong.serializer()

JesusMcCloud: There will be major upcoming changes in the CBOR codec upstream, so my take: deserialize as ULong, then do a range-check and manually throw. Valid CBOR can still overflow Kotlin's ULong, but this is something that cannot be remedied until the next major version of the CBOR codec.

JesusMcCloud: NVM. Upstream has the range checks in place.

JesusMcCloud (approved):
Just skim over it whether signed/unsigned numbers are directly (de)serialized apprpriately

PR #535, opened 2026-03-12
Change: Don't send response on user initiated signature cancellation

acrusage-iaik:
When the user cancels presentation signing, the app currently aborts the whole flow and goes to an error page.
This is because the library itself sends an error response. 

In order to improve the user presentation flow, we don't want to send an error response on user initiated signature cancellations. The user may just want to review or change selected credentials instead of aborting the whole flow.


nodh:
@codex review

nodh (approved):

PR #542, opened 2026-03-26
Remove second surrogate layer in StatusListTokenPayload serialization

n0900:
Building on #497. Removes second surrogate layer which used to differentiate between JWS and CWT serialization shape by leveraging the `preferCborLabelsOverNames` cbor serializer flag which is already required by the repository (see `Identifier.kt`)

A lot of value classes became unused as part of this change. I have removed them for now but can deprecate instead if wanted.


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialIssuer.kt
```diff
@@ -74,14 +74,16 @@ class CredentialIssuer(
         requireKeyAttestation = requireKeyAttestation,
         verifyAttestationProof = {
             val tokenStatusValid = runCatching {
-                it.payload.status?.get(JwtStatusListClaim.Specification.CLAIM_NAME)?.let { statusList ->
+                it.payload.status?.get("status_list")?.let { statusList ->
```
n0900: This change is unfortunate but the label was misused to begin with. `JwtStatusListClaim` refers to a `StatusList` element and this is a `StatusListInfo` element. For some reason they have the same serial name. 

RFC

n0900:
@codex review

openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc7519/jwt/claims/JwtSubjectClaim.kt
```diff
@@ -1,39 +0,0 @@
-package at.asitplus.wallet.lib.data.rfc7519.jwt.claims
-
-import at.asitplus.wallet.lib.data.rfc3986.UniformResourceIdentifier
-import at.asitplus.wallet.lib.data.rfc7519.primitives.StringOrUri
-import kotlinx.serialization.Serializable
```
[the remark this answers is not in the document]
nodh: I'm fine with that

openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/primitives/PositiveDurationFormatSerializer.kt
```diff
@@ -0,0 +1,10 @@
+package at.asitplus.wallet.lib.data.rfc.tokenStatusList.primitives
+
+import at.asitplus.wallet.lib.data.rfc.tokenStatusList.serialization.FormatTransformingSerializerTemplate
+import kotlinx.serialization.KSerializer
+
+object PositiveDurationFormatSerializer : KSerializer<PositiveDuration> by FormatTransformingSerializerTemplate(
```
nodh, line 10: Don't we need to register that at the class `PositiveDuration`? (in the change list one file above)

openid-data-classes/src/commonTest/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/StatusListTokenPayloadSerializationTest.kt
```diff
@@ -0,0 +1,183 @@
...
+
+        shouldThrow<SerializationException> {
+            vckJsonSerializer.decodeFromString<StatusListTokenPayload>(invalidJson.toString())
+        }
+    }
+
+    "CBOR status-list payload uses numeric labels and unsigned ttl values" {
+        val encoded = encodeCbor(StatusListTokenPayload.serializer(), statusListPayload)
+        val expectedPrefix = buildString {
+            append("A5")
```
nodh: Please add comments (or constants matching the semantic) that this is "CBOR map (5)" and so on (so one can read right away what this should mean)

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialIssuer.kt
nodh: But can we extract that in a constant?

nodh (changes requested):
Great! Some test improvements

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialIssuer.kt
n0900: Yes but placement is a bit awkward. Introduced SerialNames object

nodh (approved):

nodh:
@codex review

PR #545, opened 2026-04-21
AV blueprint

nodh:


tlenz (approved):
Looks fine, but I'm not a full expert on age verification

PR #554, opened 2026-05-19
WIA/KA Attestations 2026-05

md4096:


md4096:
Updated to ts3-wallet-unit-attestation 1.5.1

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -509,25 +519,26 @@ class OAuth2KtorClient(
...
+                    LoadInstanceAttestationInput(
+                        authorizationServer = authorizationServer,
+                        preferredClientStatusPeriod = preferredClientStatusPeriod ?: 31.days,
+                    )
+                ).getOrThrow().serialize() to
+                        BuildClientAttestationPoPJwt.invoke(
+                            signJwt = SignJwt(keyMaterial, JwsHeaderNone()),
+                            clientId = oAuth2Client.clientId,
+                            audience = authorizationServer,
+                            nonce = null
```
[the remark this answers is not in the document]
md4096: Added comment with 9989e15bc4d692b94e3fec318d98897556bcaecd

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
```diff
@@ -185,6 +179,13 @@ class OpenId4VciClient(
             )
         ).getOrThrow()
 
+        val accessToken = JwsSigned.deserialize<OpenId4VciAccessToken>(
+            it = tokenResponse.params.accessToken,
+            deserializationStrategy = OpenId4VciAccessToken.serializer(),
+        ).getOrThrow()
+
+        if(!accessToken.payload.verifyDpopThumbprint()) throw Exception("Instance attestation key not used for Dpop")
```
nodh: **P1 — Access token parsed unconditionally as JWT**

`tokenResponse.params.accessToken` is parsed as a signed JWT and immediately throws on failure. This runs on every `resumeWithAuthCode` call, including flows with opaque/reference access tokens, causing a regression: any issuer returning a non-JWT token will fail here before the credential request is even sent.

The check should be conditional — e.g. only run when the token string looks like a compact JWS, or use `getOrNull()` and skip the thumbprint check when parsing fails.

nodh (commented):
Overall this cleanup has the right direction — consolidating key material into one place is cleaner and removing the deprecated callbacks reduces API surface. However several correctness issues were introduced. The P1 items should be resolved before merging.

(Comments written by Claude Code, but confirmed by me)

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
```diff
@@ -443,6 +444,15 @@ class OpenId4VciClient(
...
             .get(insertWellKnownPath(publicContext, WellKnownPaths.OpenidConfiguration))
             .body<OAuth2AuthorizationServerMetadata>()
+
+    /**
+     * ts3-wallet-unit-attestation 1.5.1
+     * Use instance attestation key for Dpop
+     * Verify access token `cnf.jkt` matches instance attestation `cnf` key
+     */
+    private fun OpenId4VciAccessToken.verifyDpopThumbprint() =
+        this.confirmationClaim?.jsonWebKeyThumbprint == oauth2Client.keyMaterial.jsonWebKey.jwkThumbprintPlain
```
nodh, line 460: **P1 — Wrong key used for DPoP thumbprint check**

The access token's `cnf.jkt` is bound to the key used by `signDpop` (the DPoP proof key), not to `keyMaterial` (the attestation PoP key). These are the same object only in the tests because they set `dpopKeyMaterial = clientAuthKeyMaterial` — so the tests pass but the check is incorrect in general.

Any caller that supplies a custom `signDpop` with a different key will get a spurious `"Instance attestation key not used for Dpop"` failure even though the server correctly accepted the DPoP proof. The comparison needs to be against the key inside `signDpop`, not `keyMaterial`.

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -509,25 +519,26 @@ class OAuth2KtorClient(
...
-                ?: signClientAttestationPop?.invoke()?.getOrNull()?.serialize()
-        } else null
+        val (clientAttJwt, clientAttPop) = when (loadInstanceAttestation != null) {
+            true -> {
+                loadInstanceAttestation.invoke(
+                    LoadInstanceAttestationInput(
+                        authorizationServer = authorizationServer,
+                        preferredClientStatusPeriod = preferredClientStatusPeriod ?: 31.days,
+                    )
+                ).getOrThrow().serialize() to
```
nodh: **P1 — Hard failure with no recovery path**

The old code used `?.getOrNull()` so that a failed attestation callback silently skipped the attestation headers. This now calls `getOrThrow()`, and `BuildClientAttestationPoPJwt.invoke(...)` on the next line is also uncaught. If the attestation callback returns a failure or key signing throws, the entire token/PAR request propagates an unhandled exception.

`BuildClientAttestationPoPJwt.invoke(...)` should at minimum be wrapped in `catching { }.getOrThrow()`, and the error message surfaced should be descriptive.

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -509,25 +519,26 @@ class OAuth2KtorClient(
...
+                        BuildClientAttestationPoPJwt.invoke(
+                            signJwt = SignJwt(keyMaterial, JwsHeaderNone()),
+                            clientId = oAuth2Client.clientId,
+                            audience = authorizationServer,
+                            nonce = null // TODO: Add nonce after backend implementation is ready
```
nodh: **P1 - Loss of nonce support in attestation PoP**

The old `loadInstanceAttestationPop` callback received `authorizationServer`, `resourceUrl`, and `httpMethod`, which callers used to retrieve and supply a server-issued challenge/nonce. That extension point is now gone entirely. Any authorization server that requires a nonce in the client-attestation PoP will reject these requests. The TODO acknowledges this but does not note it is a functional regression. If this is intentionally deferred, there should be a tracking issue and the KDoc should warn that nonce-protected endpoints are not yet supported.

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -509,25 +519,26 @@ class OAuth2KtorClient(
...
-                )
-            )?.getOrNull()?.serialize()
-                ?: signClientAttestationPop?.invoke()?.getOrNull()?.serialize()
-        } else null
+        val (clientAttJwt, clientAttPop) = when (loadInstanceAttestation != null) {
+            true -> {
+                loadInstanceAttestation.invoke(
+                    LoadInstanceAttestationInput(
+                        authorizationServer = authorizationServer,
+                        preferredClientStatusPeriod = preferredClientStatusPeriod ?: 31.days,
```
nodh: **P2 - preferredClientStatusPeriod null semantics removed**

When the AS metadata omits `preferred_client_status_period`, the callback previously received `null` (meaning the AS has no preference, apply your own policy). Now it always receives `31.days`. Callers that distinguish null from an explicit value to decide whether to mint or reuse an attestation will silently behave incorrectly.

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -498,7 +508,7 @@ class OAuth2KtorClient(
 
     /**
      * Sets the appropriate headers when accessing a token endpoint:
-     * - loads client attestation when [loadInstanceAttestation] and [loadInstanceAttestationPop] is set
+     * - loads client attestation when [loadInstanceAttestation] is set
      * - sends a DPoP proof when [useDpop] is set
      */
     @Suppress("DEPRECATION")
```
nodh: Stale suppress annotation. The deprecated callbacks (`loadClientAttestationJwt`, `signClientAttestationPop`) that required this are now fully removed. The `@Suppress("DEPRECATION")` is dead and can be removed.

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
nodh: Fine for now

vck-openid-ktor/src/commonTest/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClientTest.kt
```diff
@@ -246,14 +233,8 @@ val OAuth2KtorClientTest by testSuite {
             ).getOrThrow()
 
             attestationInput.shouldNotBeNull().also {
```
nodh, line 231: The assertion `it.authorizationServer shouldBe authorizationService.publicContext` was removed here. The `authorizationServer` field is still present in `LoadInstanceAttestationInput` and still populated - this deletion silently reduces test coverage for a field that callers depend on for per-AS attestation selection. Please restore the assertion.

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
```diff
@@ -495,3 +505,6 @@ data class CredentialIdentifierInfo(
 private suspend inline fun <R> IntermediateResult<R>.onSuccessCredential(
     block: String.(httpResponse: HttpResponse) -> R,
 ) = onSuccess<String, R>(block)
+
+private val JsonWebKey.jwkThumbprintPlain: String
+    get() = jwkThumbprint.removePrefix("urn:ietf:params:oauth:jwk-thumbprint:sha256:")
```
nodh: Missing newline at end of file.

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
md4096: Should we move the `signDpop` from the constructor to the class as a member?
So as far as i'm aware based on the new `1.5.1` `ts3-wallet-unit-attestation`, it does not make sense to be able to separately define a `signDpop` with another keymaterial

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
md4096: Changed in 22f372605d35636f5d33f75f409530dc4f279b31

vck-openid-ktor/src/commonTest/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClientTest.kt
md4096: Changed in 22f372605d35636f5d33f75f409530dc4f279b31

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
md4096: Changed in 22f372605d35636f5d33f75f409530dc4f279b31

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
md4096: Changed in 22f372605d35636f5d33f75f409530dc4f279b31

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
md4096: Changed in 22f372605d35636f5d33f75f409530dc4f279b31 to exceptions with a description.
Or should we always use `null` in case of a failure?

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
md4096: Changed with 22f372605d35636f5d33f75f409530dc4f279b31

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
```diff
@@ -185,6 +179,13 @@ class OpenId4VciClient(
             )
         ).getOrThrow()
 
+        JwsSigned.deserialize<OpenId4VciAccessToken>(
+            it = tokenResponse.params.accessToken,
+            deserializationStrategy = OpenId4VciAccessToken.serializer(),
+        ).getOrNull()?.let {
+            if(!it.payload.verifyDpopThumbprint()) throw Exception("Instance attestation key not used for Dpop")
```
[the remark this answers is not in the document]
md4096: a96e193a394a4d0eae2b11dc67a86f8ad6466d37

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
```diff
@@ -185,6 +180,15 @@ class OpenId4VciClient(
             )
         ).getOrThrow()
 
+        if (tokenResponse.params.tokenType == TOKEN_TYPE_DPOP) {
```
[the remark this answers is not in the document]
md4096: https://datatracker.ietf.org/doc/html/rfc9449
Only `header field names` are case insensitive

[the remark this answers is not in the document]
nodh: Access token types are in fact case insensitive per https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2:
```
token_type
         REQUIRED.  The type of the token issued as described in
         [Section 7.1](https://datatracker.ietf.org/doc/html/rfc6749#section-7.1).  Value is case insensitive.
```

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -85,24 +100,19 @@ class OAuth2KtorClient(
...
-    @Deprecated("Use `loadInstanceAttestation` with `LoadInstanceAttestationInput`")
-    val loadClientAttestationJwt: (suspend () -> KmmResult<JwsSigned<JsonWebToken>>)? = null,
-    @Deprecated("Use `loadInstanceAttestationPop` with `LoadInstanceAttestationPopInput`")
-    val signClientAttestationPop: (suspend () -> KmmResult<JwsSigned<JsonWebToken>>)? = null,
+
+    /** Used to prove possession of the key material for the instance attestation **/
+    val keyMaterial: KeyMaterial = EphemeralKeyWithoutCert(),
+
+    /** Used to calculate DPoP, i.e. the key the access token and refresh token gets bound to. */
+    private val signDpop: SignJwtFun<JsonWebToken> = SignJwt(keyMaterial = keyMaterial, JwsHeaderCertOrJwk()),
```
nodh: Please do not change the order of the constructor parameters. Regarding the discussion in https://github.com/a-sit-plus/vck/pull/554#discussion_r3273841514: In the comments for the existing code we see that `signDpop` is used to bind the access token to a key and provider a PoP for that. The new parameter `keyMaterial` makes this key explicit, but below that same `keyMaterial` is also used for the client attestation PoP. This looks fine acc. to TS3. We should make this explicit in the comments.

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
nodh: Looks fine, but the issue about the tests remain, there we have
```
keyMaterial = issuerClientAuthKeyMaterial,
signDpop = SignJwt(issuerDpopKeyMaterial, JwsHeaderCertOrJwk()),
```

which looks off, even with 
```
val walletDpopKeyMaterial = walletClientAuthKeyMaterial
```

... should be simplified.

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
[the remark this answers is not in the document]
md4096: Changed with [40be3a0](https://github.com/a-sit-plus/vck/pull/554/commits/40be3a061e1f2fd1b0efe2594740d2bf368fd1bf)

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
md4096: I think removing the signDpop from the constructor would actually be a clean way. We only hand over a keyMaterial thus removing the need for separate DpoP keys (which would violate the 1.5.1 spec).

Do we have use cases for customizable DPoP Keys?

nodh: So the `keyMaterial` is then the key used for binding access tokens to and the WIA key -- fine for me, but needs a clarifying comment.

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
md4096: Deprecated signDpop with [97190ad](https://github.com/a-sit-plus/vck/pull/554/commits/97190ade72752c78332d4fd310ad44572dabcbce) as discussed.
Order adjusted because of new dependency `keyMaterial`

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OpenId4VciClient.kt
md4096: See https://github.com/a-sit-plus/vck/pull/554#discussion_r3309721304

nodh (approved):

nodh (approved):

PR #563, opened 2026-06-08
Improve readme files

nodh:


ARCHITECTURE.md
```diff
@@ -0,0 +1,247 @@
...
+VP token validation happen in `vck` and `vck-openid`.
+
+### Key Material and Crypto
+
+VC-K uses Signum for crypto, ASN.1, JOSE, and COSE types. Do not introduce parallel crypto representations unless an
+external API requires an adapter. Application-facing signing hooks usually flow through `KeyMaterial`, `Signer`, or
+specific signing/verification function types.
+
+Use `KmmResult` for public multiplatform APIs that need Swift-friendly result handling. Avoid replacing it with
+Kotlin's inline `Result` in public APIs.
```
JesusMcCloud:
```suggestion
Kotlin's inline `Result` in public APIs.  
Use `catching` as a drop-in replacement for `runCatching` to produce a `KmmResult` instead of a `Result`. Note that `catching` intentionally does not catch exceptions that cannot be handled such as out-of-heap errors. Use `catchingUnwrapped` internally whenever a `Result` is sufficient, because it also allows fatal exceptions to bubble up, but does not incur the instantiation overhead of producing a `KmmResult` instance.
```

JesusMcCloud (approved):
Pending my suggestions about KmmResult

PR #570, opened 2026-06-15
feat(oidvci): support jwk/kid key binding selection in credential request proofs

etnafed:
## Motivation

The current implementation of `WalletService` hardcodes the key binding mechanism used in the JWS header of credential request proofs, always embedding the holder's public key inline as a `jwk` header parameter. This approach does not accommodate use cases where the holder is identified by a DID, in which case the proof header should instead carry a `kid` parameter referencing a key in the holder's DID Document.

This creates potential issues:

- **Spec non-compliance**: The OID4VCI specification allows both `jwk` and `kid` as mutually exclusive key binding mechanisms in proof headers; forcing `jwk` prevents DID-based key binding
- **Interoperability limitations**: Issuers that require DID-based key binding cannot be supported with the current rigid implementation
- **Lack of extensibility**: There is no way for callers to customise the key binding strategy without modifying the library itself

Additionally, the modification comment introduced alongside this change incorrectly referenced the W3C Verifiable Credential Data Model 1.1 `iss` claim semantics, which is unrelated to the actual change. This has been corrected.

This change is necessary to ensure full compliance with the OID4VCI specification and to improve interoperability with issuers that require DID-based holder key binding.

---

### What was changed

- Added a `selectProofJwtKeyBinding` constructor parameter to `WalletService`, a suspending lambda that receives the current `KeyMaterial` and returns a `Pair<JsonWebKey?, String?>` representing the chosen key binding mechanism
- Updated `createCredentialRequestProofJwt` to invoke `selectProofJwtKeyBinding` and apply the result to the JWS header, supporting both `jwk` (inline key) and `kid` (DID URL) paths
- Added strict mutual exclusivity enforcement: returning both values non-null or both null raises an `IllegalArgumentException` at proof creation time
- Fixed an inaccurate modification comment that incorrectly cited the W3C VC Data Model 1.1 `iss` semantics; the comment now accurately describes the actual change

**Before:**
```kotlin
// JWS header always embedded the public key inline
header.copy(
    jsonWebKey = key.jsonWebKey,
    keyAttestation = keyAttestation?.jws
)
```

**After:**
```kotlin
// JWS header key binding is resolved by the caller-supplied lambda
val (jsonWebKey, keyId) = selectProofJwtKeyBinding.invoke(key)
when {
    jsonWebKey != null && keyId.isNullOrEmpty() ->
        header.copy(jsonWebKey = jsonWebKey, keyAttestation = keyAttestation?.jws)
    jsonWebKey == null && !keyId.isNullOrEmpty() ->
        header.copy(keyId = keyId, keyAttestation = keyAttestation?.jws)
    jsonWebKey != null && !keyId.isNullOrEmpty() ->
        throw IllegalArgumentException("Key binding conflict: both 'jwk' and 'kid' are set...")
    else ->
        throw IllegalArgumentException("Key binding missing: neither 'jwk' nor 'kid' is set...")
}
```

### How it works

- The default value of `selectProofJwtKeyBinding` preserves the previous behaviour: `{ key -> Pair(key.jsonWebKey, null) }`, embedding the `JsonWebKey` inline — fully backward compatible
- Callers that need DID-based key binding can supply `{ _ -> Pair(null, "did:example:holder#key-1") }`
- The lambda is invoked inside the `SignJwt` header builder, giving it access to the live `KeyMaterial` at signing time
- Mutual exclusivity of `jwk` and `kid` is enforced at runtime with clear error messages referencing the OID4VCI specification

### Key implementation details

- `selectProofJwtKeyBinding` is a `suspend` lambda, allowing asynchronous DID resolution if needed
- The parameter is intentionally a constructor argument to keep `createCredentialRequestProofJwt` signature stable
- Both `jwk` and `kid` paths correctly propagate the optional `keyAttestation` JWS compact value into the header
- The inaccurate modification comment has been replaced with an accurate description of the actual change

---

### Unit tests added

- ✅ `throw when both jwk and kid are set in selectProofJwtKeyBinding` — verifies `IllegalArgumentException` when both values are non-null
- ✅ `throw when neither jwk nor kid is set in selectProofJwtKeyBinding` — verifies `IllegalArgumentException` when both values are null
- ✅ `create proof using kid (DID URL) key binding method` — verifies successful proof creation using the `kid`-only path

### Existing tests updated

- ✅ Fixture `client` construction now explicitly passes `selectProofJwtKeyBinding = { key -> Pair(key.jsonWebKey, null) }` to document and test the default path
- ✅ `key attestation callback receives issuer preference context` — updated client construction to include explicit `selectProofJwtKeyBinding`
- ✅ All previously existing attestation tests continue to pass with the refactored proof header logic

### Regression testing

- ✅ Verified OID4VCI credential issuance flow works correctly with the default `jwk` key binding
- ✅ Verified key attestation acceptance and rejection flows are unaffected
- ✅ Verified that `requireKeyMaterialAtAttestedKeyIndex0` guard still triggers correctly
- ✅ Confirmed backward compatibility: callers that do not supply `selectProofJwtKeyBinding` retain identical behaviour to the previous implementation

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/WalletService.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/WalletService.kt
@@ -1,5 +1,26 @@
 package at.asitplus.wallet.lib.oidvci
 
+/*
+ * Software Name : VC-K
+ * SPDX-FileCopyrightText: Copyright (c) A-SIT Plus GmbH
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Modifications:
+ * - According to the W3C Verifiable Credential Data Model 1.1 https://www.w3.org/TR/vc-data-model-1.1/#jwt-encoding,
+ * "iss MUST represent the issuer property of a verifiable credential or the holder property of a verifiable presentation."
+ * So in this case the issuer should be the wallet holder, represented by it's DID.
+ * SPDX-FileCopyrightText: Copyright (c) Orange Business
+ * - Added support for configurable key binding method selection via resolveKeyBindingMethod,
+ * allowing the credential request proof JWS header to embed either an inline JsonWebKey (jwk)
+ * or a DID URL key identifier (kid), as required by the OID4VCI specification.
+ * Exactly one of jwk or kid must be provided; supplying both or neither raises an IllegalArgumentException.
+ * The default behaviour (embed jwk inline) is preserved for backward compatibility.
+ * SPDX-FileCopyrightText: Copyright (c) Orange Business
+ *
+ * This software is distributed under the Apache License 2.0,
+ * see the "LICENSE" file for more details
+ */
+
 import at.asitplus.KmmResult
 import at.asitplus.catching
 import at.asitplus.catchingUnwrapped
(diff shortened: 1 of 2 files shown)


etnafed:
Hello @nodh, @JesusMcCloud, here is a new pull request to review :)

nodh (approved):

etnafed:
@nodh Hello why did you change the Milestone from 6 to 7? Is there an issue with the PR?

nodh:
Sorry, will include in 6.0.0!

PR #575, opened 2026-06-22
DCAPI: Fix validation of unsigned requests

nodh:
to be extended after integration tests in Valera

/cc @gp-iaik


gp-iaik:
Should we remove the parsed client_id so that it cannot be used internally?

nodh:
Merged with 3ea89e46baaf2a9ea8e7f50e108cc9307b0e562d

PR #584, opened 2026-07-06
Add JVM overloads for constructors

### LINKED ISSUE #581: IssuerAgent JVM Overloads
We would need some java-friendly way to create an IssuerAgent object. We found some JVM overloads in version 5.13.0 for KeyStoreMaterial, but not for IssuerAgent, which we still have to create objects via a Kotlin wrapper class.

nodh:
Fixes #581


JesusMcCloud (approved):
We should probably let the @tlenz try it out?

PR #586, opened 2026-07-07
DCAPI: Fix serializing origin for presentation

nodh:


tlenz (approved):

gp-iaik:
I can't find it anywhere in the OpenID4VP standard that origins should be serialized (unlike ISO 18013-7). What about origins like this: `platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0`? I'd leave it to the caller to correctly serialize the expected value if required.

PR #589, opened 2026-07-08
Nonce Aware Verifier

nodh:
Part 1 of a better interface for DCAPI: Extract some nonce handling into the verifier.

Next PR: https://github.com/a-sit-plus/vck/pull/590


vck-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/openid/OpenId4VpSdJwtProtocolTest.kt
```diff
@@ -104,6 +105,27 @@ val OpenId4VpSdJwtProtocolTest by matrixSuite {
...
+                ),
+                OpenId4VpVerifier.CreationOptions.Query(it.walletUrl)
+            ).getOrThrow().url
+
+            val authnResponse = it.holderOid4vp.createAuthnResponse(authnRequest).getOrThrow()
+                .shouldBeInstanceOf<AuthenticationResponseResult.Redirect>()
+
+            it.verifierOid4vp.validateAuthnResponse(authnResponse.url).getOrThrow()
+            it.verifierOid4vp.validateAuthnResponse(authnResponse.url).isFailure shouldBe true
+        }
```
JesusMcCloud, line 127: Prime candidate to be executed concurrently like there's no tomorrow: concurrently create requests, once all are there, concurrently create responses and assert nonces are invalidated

JesusMcCloud (approved):
At your discretion, but maybe we should just tackle focused concurrent testing and stress testing in a separate PR

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/OpenId4VpVerifier.kt
```diff
@@ -526,9 +534,25 @@ class OpenId4VpVerifier @JvmOverloads constructor(
...
+                        .also { Napier.d("nonce not valid: $expectedNonce, not known to us") }
+                }
+            }
+        }
     }
 
+    private fun AuthnResponseResult.isFullyValid(): Boolean =
+        idTokenValidationResult?.isFailure != true &&
+                vpTokenValidationResult?.isFailure != true &&
+                (vpTokenValidationResult?.getOrNull()?.isFullyValid() ?: true)
```
gp-iaik, line 550: Shouldn't this better be ?: false?

nodh: Should not matter, because:
* When `vpTokenValidationResult` is an error then line 2 does evaluate to false
* When `vpTokenValidationResult` is not an error then line 3 returns `isFullyValid()`

Covered in https://github.com/a-sit-plus/vck/pull/596

PR #598, opened 2026-07-13
Fix - LoTE filtering logic

StjepanovicSrdjan:
- Fix filtering of the Trust Lists, where absence of the `ServiceTypeIdentifier` signifies that all listed trusted entity 
services are of the same type. Previous filtering logic was too strict

### DIFF
--- a/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/etsi/LoTEFilterService.kt
+++ b/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/etsi/LoTEFilterService.kt
@@ -86,17 +86,12 @@ enum class LoTEServiceType(val type: String) {
         /**
          * Resolves a raw scheme type string into a safe Enum
          */
-        fun fromSchemeType(rawSchemeType: String?): LoTEServiceType? {
-            if (rawSchemeType.isNullOrBlank()) return null
+        fun fromSchemeType(rawSchemeType: String?): LoTEServiceType {
+            if (rawSchemeType.isNullOrBlank()) return EAA
 
-            return when {
-                rawSchemeType.contains("pid", ignoreCase = true) -> PID
-                rawSchemeType.contains("mdl", ignoreCase = true) -> MDL
-                rawSchemeType.contains("wrpac", ignoreCase = true) -> WRPAC
-                rawSchemeType.contains("eaa", ignoreCase = true) -> EAA
-                rawSchemeType.contains("wallet", ignoreCase = true) -> WALLET
-                else -> null
-            }
+            return entries.firstOrNull {
+                it != EAA && rawSchemeType.contains(it.type, ignoreCase = true)
+            } ?: EAA
         }
     }
 }
--- a/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/etsi/LoTEFilterService.kt
+++ b/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/etsi/LoTEFilterService.kt
@@ -75,31 +75,28 @@ data class LoTEFilterCriteria(
     val expectedServiceType: LoTEServiceType,
 )
 
-enum class LoTEServiceType(val type: String, val fileName: String) {
-    PID("pid", "pid-providers.json"),
-    MDL("mdl", "mdl-providers.json"),
+enum class LoTEServiceType(
+    val type: String,
+    val fileName: String,
+    private val identifiers: List<String> = emptyList()
+) {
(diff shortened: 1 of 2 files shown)


vck/src/commonMain/kotlin/at/asitplus/wallet/lib/etsi/LoTEFilterService.kt
```diff
@@ -58,9 +68,35 @@ class LoTEFilterService {
...
+    WALLET("wallet");
+
+    companion object {
+        /**
+         * Resolves a raw scheme type string into a safe Enum
+         */
+        fun fromSchemeType(rawSchemeType: String?): LoTEServiceType? {
+            if (rawSchemeType.isNullOrBlank()) return null
+
+            return when {
```
nodh: I think that could be written like `entries().firstOrNull { it.type.equals(rawSchemeType, true) }` ... because why would we do a "contains" check?

vck/src/jvmTest/kotlin/at/asitplus/wallet/lib/etsi/LoTEFilterTest.kt
```diff
@@ -7029,21 +7029,21 @@ val LoTEFilterTest by matrixSuite {
...
+        val fetchUrl: String
     )
 
     testSuite("filter lote by type identifier") {
         mapOf(
-            "pidProviders" to TestData(pidProvidersFixed, "http://uri.etsi.org/19602/SvcType/PID/Issuance"),
-            "walletProviders" to TestData(walletProvidersFixed, "http://uri.etsi.org/19602/SvcType/WalletSolution/Issuance"),
-            "wrpacProviders" to TestData(wrpacProvidersFixed, "http://uri.etsi.org/19602/SvcType/WRPAC/Issuance"),
-            "mdlProviders" to TestData(mdlProvidersFixed, "http://uri.etsi.org/19602/SvcType/mDL/Issuance"),
+            "pidProviders" to TestData(pidProvidersFixed, "https://acceptance.trust.tech.ec.europa.eu/lists/eudiw/pid-providers.json"),
```
nodh: do we have constants for `pidProviders` and the other strings? Looks like we should have?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/etsi/LoTEFilterService.kt
StjepanovicSrdjan: The reason we have to use .contains() instead of .equals() is because the rawSchemeType (which comes from vcType, sdJwtType, or isoDocType) is a fully qualified URI/URN, not a simple string. For example, a PID's scheme type is usually urn:eudi:pid:1. Did add the `firstOrNull` pattern in https://github.com/a-sit-plus/vck/pull/598/commits/c4ccebb050ec5a7144acef1cfbc9f64109748576, but with contains.

vck/src/jvmTest/kotlin/at/asitplus/wallet/lib/etsi/LoTEFilterTest.kt
StjepanovicSrdjan: resolved in https://github.com/a-sit-plus/vck/pull/598/commits/5b9dd5e537a2906b1affc4435b419011222fa517

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/etsi/LoTEFilterService.kt
nodh: Okay, then we'll need to adapt two more things: Do not name the function and parameter `schemeType` but `schemeIdentifier` to align with nomenclature in VC-K. Also we would need to explicitly check if this is the PID identifier (`urn:eudi:pid:1:` for SD-JWT or `eu.europa.ec.eudi.pid.1` for MDOC, see https://github.com/eu-digital-identity-wallet/eudi-doc-attestation-rulebooks-catalog/blob/main/rulebooks/pid/pid-rulebook.md) or the mDL identifier (`org.iso.18013.5.1.mDL`) and then map it. Can it ever be a WRPAC or WALLET?

nodh (changes requested):
See comment above

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/etsi/LoTEFilterService.kt
StjepanovicSrdjan: Applied in the https://github.com/a-sit-plus/vck/pull/598/commits/827072c4edb0a2833a8ea4d89d36a3128fbae1c3.

nodh (approved):

PR #601, opened 2026-07-15
OID4VCI: Do not send proofs if issuer doesn't support any types

nodh:
Fixes https://github.com/a-sit-plus/valera/issues/455


JesusMcCloud (approved):
lgtm

Issue #529, opened 2026-02-27
Issuing: Provide certificate for identifier list


nodh:
We should probably move that out to a `StatusListIssuer` in a future PR?

_Originally posted by @nodh in https://github.com/a-sit-plus/vck/pull/519#discussion_r2845585783_

nodh:
Released in https://github.com/a-sit-plus/vck/releases/tag/7.0.0

PR #605, opened 2026-07-15
Rework Status List Agent

nodh:
Fixes https://github.com/a-sit-plus/vck/issues/529 and also the requirement to not always issue status claims for credentials, which can now be achieved by setting `IssuerAgent.statusListAgent=null`.


JesusMcCloud (approved):
LGTM

PR #608, opened 2026-07-17
Remove support for Presentation Exchange over OpenID4VP

nodh:
Remove support for Presentation Exchange over OpenID4VP, which has long been replaced with DCQL.

Next step is to prevent mapping ISO Device Requests from Proximity Flows or ISO 18013-7 Annex C over DCAPI to Presentation Exchange request. They should be supported directly in our library.

//cc @opsiv


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/OpenId4VpHolder.kt
```diff
@@ -374,7 +375,7 @@ class OpenId4VpHolder @JvmOverloads constructor(
                     )
                 }
 
-            null -> TODO()
+            else -> TODO()
```
JesusMcCloud, line 399: is CredentialPresentationRequest not sealed?

JesusMcCloud (approved):
I don't see any red flags, but ofc I also don't know if there are still remnants left that are not gone and/or not marked as deprecated

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/OpenId4VpHolder.kt
nodh: The member that is checked is optional

JesusMcCloud (approved):
LGTM

gp-iaik (approved):

PR #610, opened 2026-07-20
Decouple Holder from Presentations

nodh:
Extracts code from `HolderAgent` to keep the public API small.


JesusMcCloud (approved):
i have mostly checked for cosmetic issues, as tracing functionality that was mostly moved, is a tad tricky

PR #611, opened 2026-07-20
Prevent race conditions

nodh:


vck/src/commonTest/kotlin/at/asitplus/wallet/lib/LibraryInitializerTest.kt
```diff
@@ -39,6 +45,29 @@ private data class TestCredentialScheme(
 ) : CredentialScheme
 
 val LibraryInitializerTest by matrixSuite {
+    "serializer registration is safe when extension libraries initialize concurrently" {
+        data class ConcurrentValue(val id: String)
+
+        val prefix = uuid4().toString()
+        val start = CompletableDeferred<Unit>()
+        coroutineScope {
```
JesusMcCloud, line 51: that looks convoluted. can't we just launch a whole bunch of jobs (like 1000) that register and joinall?

JesusMcCloud (changes requested):

JesusMcCloud (approved):
i really like this!

PR #614, opened 2026-07-22
ISO: Use typed parameters

nodh:


openid-data-classes/src/commonMain/kotlin/at/asitplus/iso/IsoVersionSerializer.kt
```diff
@@ -0,0 +1,30 @@
...
+        value: Version,
+    ): Unit {
+        encoder.encodeString(value.toIsoString())
+    }
+
+    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("LooseVersion", PrimitiveKind.STRING)
+}
+
+/** Skips patch if it is 0, i.e. serializes `1.1` instead of `1.1.0` */
+fun Version.toIsoString(): String = if (patch == 0 && preRelease == null && buildMetadata == null)
```
JesusMcCloud, line 27: leniently parsed version (`strict=false`) round-trip is tested, i see. albeit implicitly. You were apparently smarter than me, when using SemVer V3

JesusMcCloud (approved):
🚢 🇮🇹

PR #616, opened 2026-07-22
Improve Java Interop

### LINKED ISSUE #606: IssuerCredentialStore.StoredCredentialReference creation in Java
Hey,

At the moment, we cannot create IssuerCredentialStore.StoredCredentialReference directly out of Java, because we haven't found a way to create the needed ULong parameter. 
Workaround for now is to write a little Kotlin wrapper just for the creation of the object, but it would be nice, if we could skip that.

nodh:
Fixes https://github.com/a-sit-plus/vck/issues/606 and other potential Java interop issues.

To quote https://kotlinlang.org/docs/unsigned-integer-types.html#non-goals
> While unsigned integers can only represent positive numbers and zero, it's not a goal to use them where application domain requires non-negative integers. For example, as a type of collection size or collection index value.

### DIFF
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/StatusListView.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/StatusListView.kt
@@ -19,9 +19,19 @@ data class StatusListView(
     fun isEmpty() = uncompressed.isEmpty()
     fun isNotEmpty() = !isEmpty()
 
+    operator fun get(index: Int) = get(index.toULong())
+
     operator fun get(index: UInt) = get(index.toULong())
+
+    operator fun get(index: Long) = get(index.toULong())
+
     operator fun get(index: ULong) = getOrNull(index)
-        ?: throw IndexOutOfBoundsException("Index $index is out of bounds, size ${uncompressed.size.toULong() * Byte.SIZE_BITS.toULong() / statusBitSize.value}")
+        ?: throw IndexOutOfBoundsException("Index $index is out of bounds, size ${size}")
+
+    private val size: ULong
+        get() = uncompressed.size.toULong() * Byte.SIZE_BITS.toULong() / statusBitSize.value
+
+    fun getOrNull(index: Long) = getOrNull(index.toULong())
 
     fun getOrNull(index: ULong): TokenStatus? {
         val tokenStatusesPerByte = Byte.SIZE_BITS.toULong() / statusBitSize.value
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/StatusListView.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/StatusListView.kt
@@ -19,11 +19,15 @@ data class StatusListView(
     fun isEmpty() = uncompressed.isEmpty()
     fun isNotEmpty() = !isEmpty()
 
-    operator fun get(index: Int) = get(index.toULong())
+    operator fun get(index: Int) = get(index.toULong().also {
+        require(index >= 0) { "index must be non-negative" }
+    })
 
     operator fun get(index: UInt) = get(index.toULong())
 
-    operator fun get(index: Long) = get(index.toULong())
+    operator fun get(index: Long) = get(index.toULong().also {
+        require(index >= 0) { "index must be non-negative" }
+    })
 
     operator fun get(index: ULong) = getOrNull(index)
         ?: throw IndexOutOfBoundsException("Index $index is out of bounds, size ${size}")
@@ -31,7 +35,9 @@ data class StatusListView(
     private val size: ULong
         get() = uncompressed.size.toULong() * Byte.SIZE_BITS.toULong() / statusBitSize.value
 
-    fun getOrNull(index: Long) = getOrNull(index.toULong())
+    fun getOrNull(index: Long) = getOrNull(index.toULong().also {
+        require(index >= 0) { "index must be non-negative" }
+    })
 
     fun getOrNull(index: ULong): TokenStatus? {
         val tokenStatusesPerByte = Byte.SIZE_BITS.toULong() / statusBitSize.value
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc3986/UniformResourceIdentifier.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc3986/UniformResourceIdentifier.kt
@@ -3,6 +3,8 @@ package at.asitplus.wallet.lib.data.rfc3986
 import io.ktor.http.*
 import kotlinx.serialization.Serializable
 import kotlin.jvm.JvmInline
+import kotlin.jvm.JvmName
+import kotlin.jvm.JvmStatic
 
 /**
  * specification: https://www.rfc-editor.org/rfc/rfc3986
@@ -19,7 +21,13 @@ import kotlin.jvm.JvmInline
 
     val string: String
         get() = url.toString()
+
+    companion object {
+        /** For JVM callers which can't access value class constructors directly */
+        @JvmStatic
+        @JvmName("fromString")
+        fun fromString(string: String) = UniformResourceIdentifier(string)
+    }
 }
 
 fun String.toUri() = UniformResourceIdentifier(this)
-
--- a/rfc3986-uri-syntax/src/commonMain/kotlin/at/asitplus/rfc3986uri/Rfc3986Authority.kt
+++ b/rfc3986-uri-syntax/src/commonMain/kotlin/at/asitplus/rfc3986uri/Rfc3986Authority.kt
@@ -8,6 +8,16 @@ data class Rfc3986Authority(
     val host: Rfc3986AuthorityHost,
     val rawPort: String?,
 ) {
+    constructor(
+        userInfo: Rfc3986UriAuthorityUserInformation?,
+        host: Rfc3986AuthorityHost,
+        port: ULong,
+    ) : this(
+        userInfo = userInfo,
+        host = host,
+        rawPort = port.toString()
+    )
+
     init {
         require(rawPort == null || rawPort.isNotEmpty() && rawPort.all { it in '0'..'9' }) {
             "port must contain decimal digits"
@@ -17,16 +27,6 @@ data class Rfc3986Authority(
     val port: ULong? = rawPort?.toULongOrNull()
 
     companion object {
-        @Deprecated(
-            "Use a String port",
-            ReplaceWith("Rfc3986Authority(userInfo, host, port.toString())"),
-        )
-        operator fun invoke(
-            userInfo: Rfc3986UriAuthorityUserInformation?,
-            host: Rfc3986AuthorityHost,
-            port: ULong,
-        ) = Rfc3986Authority(userInfo, host, port.toString())
-
         operator fun invoke(string: String): Rfc3986Authority {
             val userInfoSeparatorIndex = string.indexOf('@').takeIf {
                 it != -1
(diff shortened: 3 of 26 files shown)


JesusMcCloud:
I tend to partially disagree. Specifically: why would I want to allow for passing an illegal parameter and add range checks, that fire at runtime, when I can instead utilize the power of the compiler to prevent this outright. This is different from returning a negative number to mark a result as invalid when the domain can clearly only span positive numbers.

I know, things can get very wooly then, because you can get an index, that you might not be able to pass. Still, then it's a signal to check yourself, before you wreck yourself

nodh:
Thanks for the arguments, you're right, I've reverted most of the changes. But I'll keep the status list index change.

JesusMcCloud:
I realise that this is for java compatibility, but I'd go down a different route here, after reviewing all the long/ULong changes, because there is oh so much machinery here all over the place, for something that cannot possibly legally be signed EVER. My suggestion would rather be
* Java-Specific constructors, with different argument ordering (should be doable). make them `internal` to hide from Kotlin and add a Kdoc comment that they are Java-only.
* For functions: add Java-only functions with alternative names, mark them internal, and add `@JvmName` on them
* Add java-only getters.

I'd even argue that we might want builders/factory functions for java instead of constructors. We can still add lean upper bound checks for unsigned numeric arguments to ensure they still fit below signed `MAX_VALUE`s

I know, even more code, but it saves us from watering down our data model and function contracts.

rfc3986-uri-syntax/src/commonMain/kotlin/at/asitplus/rfc3986uri/Rfc3986Authority.kt
```diff
@@ -6,9 +6,25 @@ import kotlinx.serialization.Serializable
 data class Rfc3986Authority(
     val userInfo: Rfc3986UriAuthorityUserInformation?,
     val host: Rfc3986AuthorityHost,
-    val port: ULong?,
+    val port: String?,
```
JesusMcCloud: Why?

nodh: To keep all the port digits as parsed, and not truncate leading zeros.

JesusMcCloud: rename to `rawPort` or something like that and add a `val port: UInt get() …`

rfc3986-uri-syntax/src/commonMain/kotlin/at/asitplus/rfc3986uri/Rfc3986Authority.kt
```diff
@@ -6,9 +6,25 @@ import kotlinx.serialization.Serializable
...
+
     companion object {
+        @Deprecated(
+            "Use a String port",
+            ReplaceWith("Rfc3986Authority(userInfo, host, port.toString())"),
+        )
+        operator fun invoke(
+            userInfo: Rfc3986UriAuthorityUserInformation?,
+            host: Rfc3986AuthorityHost,
+            port: ULong,
```
JesusMcCloud: UInt should do fine (or even UShort, probably)

nodh: This is just using the old constructor syntax, but deprecating it

JesusMcCloud: but creating it programmatically should allow for passing a numeric value

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/StatusListIssuer.kt
```diff
@@ -25,18 +25,25 @@ interface StatusListIssuer : StatusIssuer, StatusProvider {
...
+     * Returns `true` if this credential has been revoked.
      */
     fun revokeCredentialByIndex(timePeriod: Int, statusListIndex: ULong): Boolean
+    
+    /**
+     * For JVM callers: Sets the status of one specific credential to
+     * [at.asitplus.wallet.lib.data.rfc.tokenStatusList.primitives.TokenStatus.Invalid].
+     * Returns `true` if this credential has been revoked.
+     */
+    fun revokeCredentialByIndex(timePeriod: Int, statusListIndex: Long): Boolean =
```
JesusMcCloud: does this not clash on the JVM targets? I'd assume this would need swapping parameters. also: make it internal and annotate with a `@JvmName`

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/agents/ReferencedTokenStore.kt
```diff
@@ -43,6 +50,12 @@ interface ReferencedTokenStore {
      */
     fun setStatus(timePeriod: Int, index: ULong, status: TokenStatus): Boolean
 
+    /**
+     * For JVM callers: Set the [status] of the referenced token with this [index] for the [timePeriod], if it exists.
+     */
+    fun setStatus(timePeriod: Int, index: Long, status: TokenStatus): Boolean =
```
JesusMcCloud: same as above

JesusMcCloud (changes requested):

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/StatusListIssuer.kt
nodh: Discussed offline: Can not be applied to interface methods

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/agents/ReferencedTokenStore.kt
nodh: Discussed offline: Can not be applied to interface methods

openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/StatusListView.kt
```diff
@@ -19,9 +19,19 @@ data class StatusListView(
     fun isEmpty() = uncompressed.isEmpty()
     fun isNotEmpty() = !isEmpty()
 
+    operator fun get(index: Int) = get(index.toULong())
+
```
JesusMcCloud: explicitly check for positivity instead of letting it silently overflow to a large positive number

openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/StatusListView.kt
```diff
@@ -19,9 +19,19 @@ data class StatusListView(
...
     operator fun get(index: UInt) = get(index.toULong())
+
+    operator fun get(index: Long) = get(index.toULong())
```
JesusMcCloud: explicitly check for positivity instead of letting it silently overflow to a large positive number

openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/StatusListView.kt
```diff
@@ -19,9 +19,19 @@ data class StatusListView(
...
+
     operator fun get(index: ULong) = getOrNull(index)
-        ?: throw IndexOutOfBoundsException("Index $index is out of bounds, size ${uncompressed.size.toULong() * Byte.SIZE_BITS.toULong() / statusBitSize.value}")
+        ?: throw IndexOutOfBoundsException("Index $index is out of bounds, size ${size}")
+
+    private val size: ULong
+        get() = uncompressed.size.toULong() * Byte.SIZE_BITS.toULong() / statusBitSize.value
+
+    fun getOrNull(index: Long) = getOrNull(index.toULong())
```
JesusMcCloud: explicitly check for positivity instead of letting it silently overflow to a large positive number

openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc3986/UniformResourceIdentifier.kt
```diff
@@ -19,7 +21,13 @@ import kotlin.jvm.JvmInline
 
     val string: String
         get() = url.toString()
+
+    companion object {
+        /** For JVM callers which can't access value class constructors directly */
+        @JvmStatic
+        @JvmName("fromString")
+        fun fromString(string: String) = UniformResourceIdentifier(string)
```
JesusMcCloud, line 29: 👍

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/StatusListIssuer.kt
```diff
@@ -25,18 +25,25 @@ interface StatusListIssuer : StatusIssuer, StatusProvider {
...
      */
     fun revokeCredentialByIndex(timePeriod: Int, statusListIndex: ULong): Boolean
 
+    /**
+     * For JVM callers: Sets the status of one specific credential to
+     * [at.asitplus.wallet.lib.data.rfc.tokenStatusList.primitives.TokenStatus.Invalid].
+     * Returns `true` if this credential has been revoked.
+     */
+    fun revokeCredentialByIndexLong(timePeriod: Int, statusListIndex: Long): Boolean =
+        revokeCredentialByIndex(timePeriod, statusListIndex.toULong())
```
JesusMcCloud: explicitly check for positivity instead of letting it silently overflow to a large positive number

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/agents/ReferencedTokenStore.kt
```diff
@@ -16,7 +16,14 @@ interface ReferencedTokenStore {
...
         val timePeriod: Int,
         val statusListIndex: ULong,
-    )
+    ) {
+        /** For JVM callers which can't access ULong directly */
+        internal constructor(
+            id: String,
+            statusListIndex: Long,
+            timePeriod: Int
+        ): this(id, timePeriod, statusListIndex.toULong())
```
JesusMcCloud: explicitly check for positivity instead of letting it silently overflow to a large positive number

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/agents/ReferencedTokenStore.kt
```diff
@@ -43,9 +50,21 @@ interface ReferencedTokenStore {
      */
     fun setStatus(timePeriod: Int, index: ULong, status: TokenStatus): Boolean
 
+    /**
+     * For JVM callers: Set the [status] of the referenced token with this [index] for the [timePeriod], if it exists.
+     */
+    fun setStatusLong(timePeriod: Int, index: Long, status: TokenStatus): Boolean =
+        setStatus(timePeriod, index.toULong(), status)
```
JesusMcCloud: explicitly check for positivity instead of letting it silently overflow to a large positive number

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/data/rfc/tokenStatusList/agents/ReferencedTokenStore.kt
```diff
@@ -43,9 +50,21 @@ interface ReferencedTokenStore {
...
+
+    /**
+     * For JVM callers: Set the [status] value of the referenced token with this [index] for the [timePeriod].
+     */
+    fun setStatusLong(timePeriod: Int, index: Long, status: Byte): Boolean =
+        setStatusLong(timePeriod, index, TokenStatus(status.toUByte()))
```
JesusMcCloud: explicitly check for positivity instead of letting it silently overflow to a large positive number

JesusMcCloud (approved):
modulo some missing safety-checks for JVM interop

PR #619, opened 2026-07-23
Remove code deprecated in 7.0.0

nodh:


JesusMcCloud (approved):
LGTM; but i have not checked it out locally and searched for leftover deprecated functionality that is to be removed

PR #626, opened 2026-07-24
Refactor Response Encryption

### LINKED ISSUE #622: Check if we can use `SessionTranscriptCalculator` here too.
Future work: Check if we can use `SessionTranscriptCalculator` here too.

_Originally posted by @nodh in https://github.com/a-sit-plus/vck/pull/587#discussion_r3636720270_

nodh:
Refactors handling response encryption by pulling extracting information about the recipient (that is the OpenID4VP Verifier) upfront into the preparation state.

Note that most changed methods are either internal themselves or in internal classes, so I did not deprecate the old methods.

Fixes also #622 by using the `SessionTranscriptCalculator` from the verifier and wallet side.


JesusMcCloud (approved):
lgtm

PR #627, opened 2026-07-24
Modular JWT

n0900:
Dependent on [Signum#463](https://github.com/a-sit-plus/signum/pull/463)

Exploratory branch to split up the `JsonWebToken` data class into smaller more specialized data classes.

Different Jwt Payloads are already used throughout the project `KeyAttestationJwt`/`KeyBindingJws`/`VerifiableCredentialJws` but only some are rolled into `JsonWebToken`.

New `JwtPayload` interface builds the backbone of all JWT claim sets, claims which are re-used (`WalletAttestationPayload`) or logically related (`ClientAttestationPayload`) are implemented as interfaces otherwise they are implemented directly as data class (`KeyAttestationPayload`).
Composition of interfaces can not be used because this would affect the serialization shape.

Looking for Feedback


openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/jwtpayload/WalletAttestationPayload.kt
```diff
@@ -0,0 +1,84 @@
...
+import kotlin.time.Instant
+
+@Deprecated("Renamed", replaceWith = ReplaceWith("WalletAttestationPayload"))
+typealias WalletAttestationClaims = WalletAttestationPayload
+
+/**
+ * Wallet Instance Attestation (WIA) as defined by
+ * [EUDI TS3](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md)
+ */
+@Deprecated("Will move into VCK next release")
```
nodh: Should not be here, since we're in VC-K already, I guess

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/jwtpayload/WalletAttestationPopPayload.kt
```diff
@@ -0,0 +1,72 @@
...
+import at.asitplus.signum.indispensable.josef.JwtClaimNames.UnregisteredClaims
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlin.time.Instant
+
+/**
+ * Wallet Instance Attestation (WIA) as defined by
+ * [EUDI TS3](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md)
+ */
+@Deprecated("Will move into VCK next release")
```
nodh: We are in VC-K

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/ClientAuthenticationService.kt
```diff
@@ -96,42 +98,25 @@ class ClientAuthenticationService @JvmOverloads constructor(
...
-        if (payload.walletSolutionCertificationInformation.isNullOrBlank()) {
-            throw InvalidClient("client attestation has no wallet_solution_certification_information")
-        }
-        val clientStatus = payload.clientStatus ?: throw InvalidClient("client attestation has no client_status")
-        if (clientStatus.expiration < (clock.now() - timeLeeway)) {
-            throw InvalidClient("client_status expiration in past: ${clientStatus.expiration}")
-        }
-        if (payload.confirmationClaim == null) {
-            throw InvalidClient("client attestation has no cnf")
+        with(payload as? WalletInstanceAttestationClaims) {
```
nodh: If `payload` is not an `WalletInstanceAttestationClaims`, something is wrong, right?

n0900: Yes we should reduce support to WIA only

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/jwtpayload/claims/ClientAttestationClaims.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Defined in [draft-ietf-oauth-attestation-based-client-auth](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-10)
+ * Section 4
+ *
+ * A JSON Web Token (JWT) generated by the
+ * Client Attester that attests to the authenticity of a Client
+ * Instance and is cryptographically bound to a key managed by that
+ * Client Instance.  A Client Attestation JWT may additionally convey
+ * information about the integrity or state of the Client Instance.
+ *
+ * Not to be confused with the related proofs of possession [ClientAttestationPopClaims]
```
nodh, line 17: But then why does this interface implement `ProofOfPossessionClaims`?

n0900: ah this is bad wording. The `ClientAttestationJwt` is a proof of possession but sent from the `ClientAttester` to the client and defined in [Section 4](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-10#section-4).

Confusingly the almost identically named `ClientAttestationPopJwt` has a completely different use-case and is sent along side the `ClientAttestationJwt` to the authorization service which is defined in [Section 5.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-10#section-5.1)
Maybe
```
* Not to be confused with the related proof of possession defined in Section 5.1 [ClientAttestationPopClaims]
```

PR #630, opened 2026-08-03
Implement RFC 9701 JWT response representation and HTTP content negotiation

n0900:
Implements correct RFC9701 wire format (sections 4, 5, 8.1, 10.3, and 10.4). Does not modify existing flows using this format.
`TokenIntrospectionJwt` defines new content type and content negotiation is now handled via `Accept` Header.

Accept semantics implemented as in [RFC9110](https://datatracker.ietf.org/doc/html/rfc9110#section-12.5.1)

### DIFF
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
@@ -1,7 +1,7 @@
 package at.asitplus.openid
 
 import io.ktor.http.ContentType
-import io.ktor.http.parseHeaderValue
+import io.ktor.http.parseAndSortContentTypeHeader
 
 sealed interface TokenIntrospectionResponse {
 
@@ -11,28 +11,31 @@ sealed interface TokenIntrospectionResponse {
             acceptHeader: String?,
             defaultResponseFormat: ContentType,
         ): ContentType {
-            val entries = parseHeaderValue(acceptHeader ?: ContentType.Any.toString()).map {
-                val mediaRange = it.params
-                    .filterNot { parameter -> parameter.name.equals("q", ignoreCase = true) }
-                    .fold(ContentType.parse(it.value)) { contentType, parameter ->
-                        contentType.withParameter(parameter.name, parameter.value)
-                    }
-                AcceptHeaderEntry(
-                    mediaRange = mediaRange,
-                    quality = it.quality,
-                )
+            val entries = parseAndSortContentTypeHeader(acceptHeader ?: ContentType.Any.toString()).map {
+                val parsedMediaRange = ContentType.parse(it.value)
+                ContentType(
+                    contentType = parsedMediaRange.contentType,
+                    contentSubtype = parsedMediaRange.contentSubtype,
+                    parameters = it.params.filterNot { parameter ->
+                        parameter.name.equals("q", ignoreCase = true)
+                    },
+                ) to it.quality
             }
 
             val candidatesWithEffectiveQuality = listOf(
                 TokenIntrospectionResponseJwt.contentType,
                 TokenIntrospectionResponseJson.contentType,
             ).mapNotNull { candidate ->
-                entries.filter { candidate.match(it.mediaRange) }
-                    .maxWithOrNull(
-                        compareBy<AcceptHeaderEntry> { it.mediaRange.typeSpecificity }
-                            .thenBy { it.mediaRange.parameters.size }
-                    )
-                    ?.quality
+                entries.filter { candidate.match(it.first) }
+                    .reduceOrNull { current, next ->
+                        val currentMediaRange = current.first
+                        val nextMediaRange = next.first
+                        if (
+                            nextMediaRange.match(currentMediaRange) &&
+                            !currentMediaRange.match(nextMediaRange)
+                        ) next else current
+                    }
+                    ?.second
                     ?.let { candidate to it }
             }
 
@@ -50,18 +53,6 @@ sealed interface TokenIntrospectionResponse {
             return defaultResponseFormat.takeIf { it in preferredCandidates }
                 ?: preferredCandidates.first()
         }
-
-        private data class AcceptHeaderEntry(
-            val mediaRange: ContentType,
-            val quality: Double,
-        )
-
-        private val ContentType.typeSpecificity: Int
-            get() = when {
-                contentType == "*" -> 0
-                contentSubtype == "*" -> 1
-                else -> 2
-            }
 
     }
 }
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
@@ -1,5 +1,7 @@
 package at.asitplus.openid
 
+import at.asitplus.KmmResult
+import at.asitplus.catching
 import io.ktor.http.ContentType
 import io.ktor.http.parseAndSortContentTypeHeader
 
@@ -10,7 +12,7 @@ sealed interface TokenIntrospectionResponse {
         fun parseAcceptHeader(
             acceptHeader: String?,
             defaultResponseFormat: ContentType,
-        ): ContentType {
+        ): KmmResult<ContentType> = catching {
             val entries = parseAndSortContentTypeHeader(acceptHeader ?: ContentType.Any.toString()).map {
                 ContentType.parse(it.value) to it.quality
             }
@@ -43,7 +45,7 @@ sealed interface TokenIntrospectionResponse {
                 .filter { it.second == highestQuality }
                 .map { it.first }
 
-            return defaultResponseFormat.takeIf { it in preferredCandidates }
+            defaultResponseFormat.takeIf { it in preferredCandidates }
                 ?: preferredCandidates.first()
         }
 
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
@@ -1,0 +1,3 @@
+package at.asitplus.openid
+
+sealed interface TokenIntrospectionResponse
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
@@ -12,15 +12,13 @@ sealed interface TokenIntrospectionResponse {
             defaultResponseFormat: ContentType,
         ): ContentType {
             val entries = parseHeaderValue(acceptHeader ?: ContentType.Any.toString()).map {
-                val parsedContentType = ContentType.parse(it.value)
+                val mediaRange = it.params
+                    .filterNot { parameter -> parameter.name.equals("q", ignoreCase = true) }
+                    .fold(ContentType.parse(it.value)) { contentType, parameter ->
+                        contentType.withParameter(parameter.name, parameter.value)
+                    }
                 AcceptHeaderEntry(
-                    mediaRange = ContentType(
-                        contentType = parsedContentType.contentType,
-                        contentSubtype = parsedContentType.contentSubtype,
-                        parameters = it.params.takeWhile { parameter ->
-                            !parameter.name.equals("q", ignoreCase = true)
-                        },
-                    ),
+                    mediaRange = mediaRange,
                     quality = it.quality,
                 )
             }
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
@@ -14,7 +14,12 @@ sealed interface TokenIntrospectionResponse {
             defaultResponseFormat: ContentType,
         ): KmmResult<ContentType> = catching {
             val entries = parseAndSortContentTypeHeader(acceptHeader ?: ContentType.Any.toString()).map {
-                ContentType.parse(it.value) to it.quality
+                val parsedContentType = ContentType.parse(it.value)
+                ContentType(
+                    parsedContentType.contentType,
+                    parsedContentType.contentSubtype,
+                    it.params.takeWhile { parameter -> !parameter.name.equals("q", ignoreCase = true) },
+                ) to it.quality
             }
 
             val candidatesWithEffectiveQuality = listOf(
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
@@ -1,3 +1,69 @@
 package at.asitplus.openid
 
-sealed interface TokenIntrospectionResponse
+import io.ktor.http.ContentType
+import io.ktor.http.parseHeaderValue
+
+sealed interface TokenIntrospectionResponse {
+
+    companion object {
+        /** Selects the supported response format with the highest effective quality. */
+        fun parseAcceptHeader(
+            acceptHeader: String?,
+            defaultResponseFormat: ContentType,
+        ): ContentType {
+            val entries = parseHeaderValue(acceptHeader ?: ContentType.Any.toString()).map {
+                val parsedContentType = ContentType.parse(it.value)
+                AcceptHeaderEntry(
+                    mediaRange = ContentType(
+                        contentType = parsedContentType.contentType,
+                        contentSubtype = parsedContentType.contentSubtype,
+                        parameters = it.params.takeWhile { parameter ->
+                            !parameter.name.equals("q", ignoreCase = true)
+                        },
+                    ),
+                    quality = it.quality,
+                )
+            }
+
+            val candidatesWithEffectiveQuality = listOf(
+                TokenIntrospectionResponseJwt.contentType,
+                TokenIntrospectionResponseJson.contentType,
+            ).mapNotNull { candidate ->
+                entries.filter { candidate.match(it.mediaRange) }
+                    .maxWithOrNull(
+                        compareBy<AcceptHeaderEntry> { it.mediaRange.typeSpecificity }
+                            .thenBy { it.mediaRange.parameters.size }
+                    )
+                    ?.quality
+                    ?.let { candidate to it }
+            }
+
+            val highestQuality = candidatesWithEffectiveQuality
+                .maxOfOrNull { it.second }
+                ?.takeIf { it > 0.0 }
+                ?: throw IllegalArgumentException(
+                    "The Accept header does not contain a supported response format."
+                )
+
+            val preferredCandidates = candidatesWithEffectiveQuality
+                .filter { it.second == highestQuality }
+                .map { it.first }
+
+            return defaultResponseFormat.takeIf { it in preferredCandidates }
+                ?: preferredCandidates.first()
+        }
+
+        private data class AcceptHeaderEntry(
+            val mediaRange: ContentType,
+            val quality: Double,
+        )
+
+        private val ContentType.typeSpecificity: Int
+            get() = when {
+                contentType == "*" -> 0
+                contentSubtype == "*" -> 1
+                else -> 2
+            }
+
+    }
+}
--- a/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
+++ b/openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
@@ -12,14 +12,7 @@ sealed interface TokenIntrospectionResponse {
             defaultResponseFormat: ContentType,
         ): ContentType {
             val entries = parseAndSortContentTypeHeader(acceptHeader ?: ContentType.Any.toString()).map {
-                val parsedMediaRange = ContentType.parse(it.value)
-                ContentType(
-                    contentType = parsedMediaRange.contentType,
-                    contentSubtype = parsedMediaRange.contentSubtype,
-                    parameters = it.params.filterNot { parameter ->
-                        parameter.name.equals("q", ignoreCase = true)
-                    },
-                ) to it.quality
+                ContentType.parse(it.value) to it.quality
             }
 
             val candidatesWithEffectiveQuality = listOf(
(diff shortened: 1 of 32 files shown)


openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
```diff
@@ -1,128 +1,51 @@
...
-    val jwt: String,
-) : TokenIntrospectionResult
+import io.ktor.http.ContentType
+import io.ktor.http.parseAndSortContentTypeHeader
+
+sealed interface TokenIntrospectionResponse {
+
+    companion object {
+        /** Selects the supported response format with the highest effective quality. */
+        fun parseAcceptHeader(
```
nodh, line 10: This method looks rather large, I'm surprised there is no functionality for that in ktor, but I'm also sure you've checked that.

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponseJwtPayload.kt
```diff
@@ -0,0 +1,37 @@
+package at.asitplus.openid
+
+import at.asitplus.signum.indispensable.io.InstantLongSerializer
+import at.asitplus.signum.indispensable.io.TransformingSerializerTemplate
```
nodh, line 4: Several unused imports here

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponseJson.kt
```diff
@@ -0,0 +1,133 @@
+package at.asitplus.openid
+
+import at.asitplus.signum.indispensable.io.InstantLongSerializer
+import at.asitplus.wallet.lib.data.IntrospectionJwt
```
nodh, line 4: unused import

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -428,19 +429,21 @@ class OAuth2KtorClient(
...
     suspend fun callTokenIntrospection(
         oauthMetadata: OAuth2AuthorizationServerMetadata,
+        responseFormat: ContentType,
         request: TokenIntrospectionRequest,
         token: String,
         popAudience: String,
         retryCount: Int = 0,
         issuerMetadata: IssuerMetadata? = null,
-    ): TokenIntrospectionResponse = oauthMetadata.introspectionEndpoint?.let { url ->
+    ): TokenIntrospectionResponseJson = oauthMetadata.introspectionEndpoint?.let { url ->
```
nodh, line 442: This change of the return type should be documented in the changelog ... but shouldn't this return the super type `TokenIntrospectionResponse`?

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -428,19 +429,21 @@ class OAuth2KtorClient(
 
     /**
      * Calls the token introspection endpoint ([OAuth2AuthorizationServerMetadata.introspectionEndpoint])
-     * to check whether the given token is active, returns [TokenInfo] on success, otherwise throws [InvalidToken].
+     * to check whether the given token is active, returns [TokenIntrospectionResponseJson] on success, otherwise throws [InvalidToken].
      */
     suspend fun callTokenIntrospection(
         oauthMetadata: OAuth2AuthorizationServerMetadata,
+        responseFormat: ContentType,
```
nodh, line 436: This new parameter should be documented in the changelog, and the old method signature should forward to the new one.
Please also move into a position after `request`, and name it something that matches the intent of the "accept" header? It is not the format of the response, but the requested format (or a better name).

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -598,29 +605,19 @@ data class TokenResponseWithDpopNonce(
...
-): TokenIntrospectionResponse = catchingUnwrapped {
-    if (requestedResponseFormat == TokenIntrospectionRequest.ResponseFormat.JWT) {
-        parseJwt(body, verifyTokenIntrospectionJwt)
-    } else {
-        catchingUnwrapped {
-            joseCompliantSerializer.decodeFromString(TokenIntrospectionResponse.serializer(), body)
-        }.getOrElse {
-            parseJwt(body, verifyTokenIntrospectionJwt)
-        }
+    responseFormat: ContentType,
```
nodh, line 608: shouldn't we read the content type the server sends when parsing the response? In my opinion we should not fail here when we request JWT, but the server doesn't support it.

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/AuthorizationService.kt
```diff
@@ -85,12 +85,15 @@ interface AuthorizationService {
...
     /**
      * [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662): OAuth 2.0 Token Introspection
+     * [RFC 9701](https://datatracker.ietf.org/doc/rfc9701/): JWT Response for OAuth 2.0 Token Introspection
      *
+     * @param acceptHeader as sent from the client as HTTP `Accept` header, or `null` if the header was omitted
      * @param request as sent from the client as form POST
      * @param httpRequest information about the HTTP request from the client, to validate authentication
      */
     suspend fun tokenIntrospection(
+        acceptHeader: String?,
```
nodh, line 95: Needs a changelog entry and the old method deprecated and forwarding to the new method. And please keep `request` at first position, since `acceptHeader` is optional, and probably should have a default value.

vck-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidvci/OidvciCodeFlowTest.kt
```diff
@@ -58,6 +58,7 @@ import io.kotest.matchers.nulls.shouldNotBeNull
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.string.shouldNotBeEmpty
 import io.kotest.matchers.types.shouldBeInstanceOf
+import io.ktor.http.*
```
nodh, line 61: not needed

vck-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidvci/OidvciAttestationTest.kt
```diff
@@ -66,6 +66,7 @@ import io.kotest.matchers.nulls.shouldNotBeNull
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.string.shouldContain
 import io.kotest.matchers.types.shouldBeInstanceOf
+import io.ktor.http.*
```
nodh, line 69: not needed

nodh (changes requested):

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
n0900: Well after verifying the JWT itself I suppose only the `TokenIntrospection` part is worth keeping? Saves us from case-distinctions downstream.

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/AuthorizationService.kt
n0900: Changelog: Done
Deprecation: Done
Default value: By default the `accept` header may be absent which is then treated as `*/*` so `null` is not neutral. Because it is important i want the consumer to hand it over explicitly

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
n0900: Ktor as a server does via the content negotiation plugin but it doesnt expose a traditional function as far as i can see. The only thing me (and codex) were able to find was `parseAndSortContentTypeHeader` but then you cannot select the top results because of the `q` parameter

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
n0900: In that case we should refactor this to allow arbitrary `accept` header formats

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/AuthorizationService.kt
```diff
@@ -83,14 +84,25 @@ interface AuthorizationService {
...
      * [RFC7662](https://datatracker.ietf.org/doc/html/rfc7662): OAuth 2.0 Token Introspection
+     * [RFC 9701](https://datatracker.ietf.org/doc/rfc9701/): JWT Response for OAuth 2.0 Token Introspection
      *
      * @param request as sent from the client as form POST
+     * @param acceptHeader as sent from the client as HTTP `Accept` header, or `null` if the header was omitted
      * @param httpRequest information about the HTTP request from the client, to validate authentication
      */
     suspend fun tokenIntrospection(
         request: TokenIntrospectionRequest,
+        acceptHeader: String?,
```
nodh: Should probably be put into `RequestInfo`

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -598,29 +630,21 @@ data class TokenResponseWithDpopNonce(
...
-        }
+    responseContentType: ContentType?,
+    verifyTokenIntrospectionJwt: suspend (JwsCompactTyped<TokenIntrospectionResponseJwtPayload>) -> Boolean,
+): TokenIntrospectionResponseJson = catchingUnwrapped {
+    when (responseContentType?.withoutParameters()) {
+        TokenIntrospectionResponseJson.contentType ->
+            joseCompliantSerializer.decodeFromString<TokenIntrospectionResponseJson>(body)
+
+        TokenIntrospectionResponseJwt.contentType -> JwsCompactTyped<TokenIntrospectionResponseJwtPayload>(body).apply {
+            require(verifyTokenIntrospectionJwt(this)) { "Token introspection JWT validation failed" }
```
nodh, line 641: but here the signature is never checked, right?

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/SimpleAuthorizationService.kt
```diff
@@ -654,34 +668,44 @@ class SimpleAuthorizationService @JvmOverloads constructor(
...
     override suspend fun tokenIntrospection(
         request: TokenIntrospectionRequest,
+        acceptHeader: String?,
         httpRequest: RequestInfo?,
-    ): KmmResult<TokenIntrospectionResult> = catching {
-        // TODO Which client_id to pass?
-        clientAuthenticationService.authenticateClient(httpRequest, null)
+    ): KmmResult<TokenIntrospectionResponse> = catching {
+        val authenticatedResourceServerId =
+            clientAuthenticationService.authenticateClientAndGetIdentity(httpRequest, null)
```
nodh, line 674: I'm working on a refactoring of binding client identities to request ... can we keep the code here as it is for now please?

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
```diff
@@ -1,128 +1,51 @@
...
+                        ) next else current
+                    }
+                    ?.second
+                    ?.let { candidate to it }
+            }
+
+            val highestQuality = candidatesWithEffectiveQuality
+                .maxOfOrNull { it.second }
+                ?.takeIf { it > 0.0 }
+                ?: throw IllegalArgumentException(
```
nodh, line 40: these exception types should be mapped to an `OAuth2Exception` when called from `SimpleAuthorizationService` (or return `KmmResult` here?)

openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/MediaTypes.kt
```diff
@@ -24,5 +26,16 @@ data object MediaTypes {
...
     }
 }
+
+/**
+ * KTOR integration: Receiver parameter for namespace
+ * See [MediaTypes.Application.INTROSPECTION_JWT]
+ */
+@Suppress("UnusedReceiverParameter")
+val ContentType.Application.IntrospectionJwt: ContentType
+    get() = ContentType.parse(INTROSPECTION_JWT)
```
nodh, line 41: better a hard coded value instead of a getter so it's not computed on every access

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
```diff
@@ -426,21 +427,41 @@ class OAuth2KtorClient(
...
         popAudience: String,
+        requestedResponseFormats: List<ContentType>,
         retryCount: Int = 0,
         issuerMetadata: IssuerMetadata? = null,
-    ): TokenIntrospectionResponse = oauthMetadata.introspectionEndpoint?.let { url ->
+    ): TokenIntrospectionResponseJson = oauthMetadata.introspectionEndpoint?.let { url ->
         Napier.i("callTokenIntrospection: $url with $request")
         val response = try {
             client.request {
+                requestedResponseFormats.forEach { accept(it) }
```
nodh, line 464: This emits several `Accept` headers, but I thought that there should be only one header, and then a comma separated list: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/SimpleAuthorizationService.kt
nodh: See https://github.com/a-sit-plus/vck/pull/646

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
n0900: From the description of `accept`: "Appends the Accept header with a specific contentType." So it should not add a new header each time

nodh: My tests say otherwise, also `append` refers to the list of headers here, see https://github.com/ktorio/ktor/blob/main/ktor-client/ktor-client-core/common/src/io/ktor/client/request/utils.kt#L87 and https://github.com/ktorio/ktor/blob/main/ktor-utils/common/src/io/ktor/util/StringValues.kt#L401

n0900: mb misread the docs

JesusMcCloud:
@n0900 request s review from me, once everything is addressed

openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/MediaTypes.kt
n0900: If we remove the getter we get "Extension property cannot be initialized because it has no backing field."
how big of an issue is this? I think the unified namespace is quite valuable

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
n0900: After reviewing this further multiple accept headers is allowed according to [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-field-lines-and-combined-fi) and seemingly also the more stable way of working with KTOR.

Their ContentNegotiation plugin explicitly uses the same method in [L187-196](https://github.com/ktorio/ktor/blob/5ba9d6fdf1ea9acfac5de67e7fe5a072639eac64/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt#L187).
This adds a new `Accept` header for each configured serial format if it is not already appended.

Additionally I found a bug while looking through this when using combined headers
```
val ContentNegotiationAcceptHeaderTest by matrixSuite {
    test("JSON content negotiation adds a second Accept header") {
        val mockEngine = MockEngine { respondOk() }

        HttpClient(mockEngine) {
            install(ContentNegotiation) { json() }
        }.use { client ->
            client.get("https://example.com") {
                header(HttpHeaders.Accept, "text/plain;q=0.5")
            }
        }

        mockEngine.requestHistory.single().headers.getAll(HttpHeaders.Accept) shouldBe listOf(
            "text/plain;q=0.5",
            "application/json",
        )
    }

    test("JSON content negotiation adds implicit q=1 despite q=0 in a combined Accept header") {
        val combinedAcceptHeader = "application/json;q=0, text/plain;q=0.5"
        val mockEngine = MockEngine { respondOk() }

        HttpClient(mockEngine) {
            install(ContentNegotiation) { json() }
        }.use { client ->
            client.get("https://example.com") {
                header(HttpHeaders.Accept, combinedAcceptHeader)
            }
        }

        // Ktor's duplicate check uses ContentType.parse, which retains only the last media range.
        val acceptHeaders = mockEngine.requestHistory.single().headers.getAll(HttpHeaders.Accept).shouldNotBeNull()
        acceptHeaders shouldBe listOf(
            combinedAcceptHeader,
            "application/json",
        )

        // The explicit JSON exclusion conflicts with the implicit q=1 added by ContentNegotiation.
        parseHeaderValue(acceptHeaders.joinToString(", "))
            .filter { it.value == "application/json" }
            .map { it.quality } shouldBe listOf(0.0, 1.0)
    }
}

```


Edit: Fixed in upcoming KTOR version https://github.com/ktorio/ktor/pull/5862

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/OAuth2KtorClient.kt
n0900: it does whatever `verifyTokenIntrospectionJwt` does. Currently it just points to `true` yes.

n0900:
Current gaps out of scope for this branch which still need to be addressed (Codex audit):

  - JSON introspection succeeds without an authenticated caller. RFC 9701 requires authentication for every introspection request and HTTP 400 for unauthenticated requests.
  - The authenticated resource server is not checked against the introspected token’s intended audience or data-access policy.
  - OAuth2KtorClient accepts syntactically valid JWT responses by default without verifying the signature, typ, iss, aud, or iat.
  - Introspection requests select authentication using token_endpoint_auth_methods_supported instead of introspection_endpoint_auth_methods_supported.
  - RFC 9701 response-signing and encryption metadata fields are not modeled or published.

n0900:
@codex review

vck-openid-ktor/src/commonMain/kotlin/at/asitplus/wallet/lib/ktor/openid/RemoteOAuth2AuthorizationServerAdapter.kt
```diff
@@ -60,6 +62,9 @@ class RemoteOAuth2AuthorizationServerAdapter(
     val internalTokenVerificationService: TokenVerificationService,
     /** Used to provide DPoP nonces for credential requests, which will be verified by [internalTokenVerificationService]. */
     val dpopNonceService: NonceService = DefaultNonceService(),
+    /** Response media ranges sent to the remote token introspection endpoint. */
+    private val tokenIntrospectionResponseFormats: List<ContentType> =
+        listOf(TokenIntrospectionResponseJwt.contentType),
```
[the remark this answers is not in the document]
n0900: Ktor ContentNegotiation should automatically add `Json` if installed

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
```diff
@@ -1,128 +1,58 @@
...
+            acceptHeader: String?,
+            defaultResponseFormat: ContentType,
+        ): KmmResult<ContentType> = catching {
+            val entries = parseAndSortContentTypeHeader(acceptHeader ?: ContentType.Any.toString()).map {
+                val parsedContentType = ContentType.parse(it.value)
+                ContentType(
+                    parsedContentType.contentType,
+                    parsedContentType.contentSubtype,
+                    it.params.takeWhile { parameter -> !parameter.name.equals("q", ignoreCase = true) },
+                ) to it.quality
```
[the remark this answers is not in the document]
n0900: Fixed in https://github.com/ktorio/ktor/pull/5794

openid-data-classes/src/commonMain/kotlin/at/asitplus/wallet/lib/data/MediaTypes.kt
nodh: Oh yeah, then keep it!

nodh (commented):
See open comment threads

openid-data-classes/src/commonMain/kotlin/at/asitplus/openid/TokenIntrospectionResponse.kt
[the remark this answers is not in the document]
JesusMcCloud: Grrrr! Ktor 3.6.0 just released AFTER I released new conventions. Anyhow, we can remove this and update to Ktor 3.6.0 (unless 3.6.0 introduces regressions)

JesusMcCloud (changes requested):
I think we should try and see if Ktor 3.6.0 *just works*, so we can remove

PR #632, opened 2026-08-04
COSE: Match fully-specified with non-fully specified algorithms

nodh:
Fixes https://github.com/a-sit-plus/valera/issues/527


JesusMcCloud (approved):
That is the correct solution to the problem!  
Signum should map strict, not fuzzy.

PR #634, opened 2026-08-04
ISO: ZKP presentation generation

opsiv:
- Introduce ZKP backend interface for generating and verifying IsoMdocProofs
- Add registry for ZKP backends
- Add engine that routes ZKP requests to the corresponding backend
- add glue logic for presentation generation using ZKP backends

### DIFF
--- a/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/PresentationDataClasses.kt
+++ b/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/PresentationDataClasses.kt
@@ -46,13 +46,16 @@ data class PresentationRequestParameters(
     val returnOneDeviceResponse: Boolean = false,
 
     /**
-     * Optional callback to calculate the ISO mDoc Session Transcript.
+     * Callback to calculate the ISO mDoc Session Transcript.
      *
      * This is a callback because calculating the Session Transcript might fail for non-mDoc presentations.
      * By using a callback, we ensure that it's only calculated when an ISO mDoc is actually
-     * part of the presentation, avoiding unnecessary failures for SD-JWT or W3C VC presentations.
+     * part of the presentation, avoiding unnecessary failures, e.g., for SD-JWT presentations.
      */
-    val calcSessionTranscript: (suspend () -> SessionTranscript)? = null,
+    val calcIsoSessionTranscript: (suspend () -> SessionTranscript) = {
+        throw IllegalStateException("Session transcript calculation callback was not provided. " +
+                "This is required for ISO mDoc presentations.")
+    },
 ) {
     /**
      * According to OID4VP 1.0 B3.3.1 every TransactionData entry may define different Digest algorithms
--- a/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/PresentationDataClasses.kt
+++ b/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/PresentationDataClasses.kt
@@ -2,6 +2,7 @@ package at.asitplus.wallet.lib.agent
 
 import at.asitplus.dif.PresentationSubmission
 import at.asitplus.iso.DeviceNameSpaces
+import at.asitplus.iso.SessionTranscript
 import at.asitplus.jsonpath.core.NormalizedJsonPath
 import at.asitplus.openid.TransactionDataBase64Url
 import at.asitplus.openid.dcql.DCQLCredentialQueryIdentifier
@@ -42,7 +43,16 @@ data class PresentationRequestParameters(
         "Only applies to deprecated Presentation Exchange. DCQL uses `DCQLCredentialQuery.multiple`; " +
                 "ISO Device Retrieval always creates one DeviceResponse."
     )
-    val returnOneDeviceResponse: Boolean = false
+    val returnOneDeviceResponse: Boolean = false,
+
+    /**
+     * Optional callback to calculate the ISO mDoc Session Transcript.
+     *
+     * This is a callback because calculating the Session Transcript might fail for non-mDoc presentations.
+     * By using a callback, we ensure that it's only calculated when an ISO mDoc is actually
+     * part of the presentation, avoiding unnecessary failures for SD-JWT or W3C VC presentations.
+     */
+    val calcSessionTranscript: (suspend () -> SessionTranscript)? = null,
 ) {
     /**
      * According to OID4VP 1.0 B3.3.1 every TransactionData entry may define different Digest algorithms
--- a/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/VerifiablePresentationFactory.kt
+++ b/vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/VerifiablePresentationFactory.kt
@@ -23,6 +23,7 @@ import at.asitplus.iso.DeviceSigned
 import at.asitplus.iso.Document
 import at.asitplus.iso.IssuerSigned
 import at.asitplus.iso.IssuerSignedItem
+import at.asitplus.iso.ZkDocument
 import at.asitplus.jsonpath.core.NormalizedJsonPath
 import at.asitplus.jsonpath.core.NormalizedJsonPathSegment
 import at.asitplus.openid.dcql.DCQLClaimsQueryResult
@@ -63,7 +64,7 @@ class VerifiablePresentationFactory(
         SignJwt(keyMaterial, JwsHeaderCertOrJwk()),
     private val signKeyBinding: SignJwtFun<KeyBindingJws> =
         SignJwt(keyMaterial, JwsHeaderNone()),
-    private val mdocZkEngine: IsoMdocZkEngine = IsoMdocZkEngine.Default
+    private val mdocZkEngine: IsoMdocZkEngine = IsoMdocZkEngine()
 ) {
     @Deprecated("Use createVerifiablePresentation(request, isoPresentationParameters) instead")
     suspend fun createVerifiablePresentation(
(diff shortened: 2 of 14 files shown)


vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/PresentationDataClasses.kt
```diff
@@ -42,7 +43,16 @@ data class PresentationRequestParameters(
...
+    val returnOneDeviceResponse: Boolean = false,
+
+    /**
+     * Optional callback to calculate the ISO mDoc Session Transcript.
+     *
+     * This is a callback because calculating the Session Transcript might fail for non-mDoc presentations.
+     * By using a callback, we ensure that it's only calculated when an ISO mDoc is actually
+     * part of the presentation, avoiding unnecessary failures for SD-JWT or W3C VC presentations.
+     */
+    val calcSessionTranscript: (suspend () -> SessionTranscript)? = null,
```
nodh: Why can't we do that with `calcIsoDeviceSignaturePlain`, or why do we need both parameters now?

vck/src/commonTest/kotlin/at/asitplus/wallet/lib/agent/ValidatorVpTest.kt
```diff
@@ -100,7 +101,10 @@ val ValidatorVpTest by matrixSuite {
                 val validator = validator
 
                 val holder = holder
-                val verifiablePresentationFactory = VerifiablePresentationFactory(holderKeyMaterial)
+                val verifiablePresentationFactory = VerifiablePresentationFactory(
+                    keyMaterial = holderKeyMaterial,
+                    mdocZkEngine = IsoMdocZkEngine(IsoMdocZkBackendRegistry())
```
nodh: do we need this parameter here? I don't see any other changes in this file to produce ZKPs

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/zk/iso/IsoMdocZkBackend.kt
```diff
@@ -0,0 +1,36 @@
...
+import at.asitplus.KmmResult
+import at.asitplus.iso.SessionTranscript
+import at.asitplus.iso.ZkDocument
+import at.asitplus.iso.ZkSystem
+import at.asitplus.jsonpath.core.NormalizedJsonPath
+import at.asitplus.wallet.lib.agent.PresentationRequestParameters
+import at.asitplus.wallet.lib.agent.SubjectCredentialStore
+import kotlinx.serialization.KSerializer
+
+interface IsoMdocZkBackend {
```
nodh, line 22: Please add short KDocs on what to implement for the methods, or this interface is general

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/VerifiablePresentationFactory.kt
```diff
@@ -185,21 +187,57 @@ class VerifiablePresentationFactory(
...
+
+        return CreatePresentationResult.DeviceResponse(
+            deviceResponse = DeviceResponse(
+                parsedVersion = Version(1, 0),
+                documents = plainDocMap.values.mapNotNull { it.getOrNull() }.toTypedArray(),
+                zkDocuments = zkDocMap.values.mapNotNull { it.getOrNull()?.toZkDocument() }.toTypedArray(),
+                status = 0U,
+            ),
+        )
+    }
```
nodh, line 222: This code is much easier to read imho:

```suggestion
        val plainDocuments = mutableListOf<Document>()
        val zkDocuments = mutableListOf<ZkDocument>()
        isoPresentationParameters.forEach { param ->
            if (param.zkMetadata is ZkMetadata.IsoMdocZk) {
                val zkMetadata = param.zkMetadata as ZkMetadata.IsoMdocZk
                mdocZkEngine.generate(request, param).fold(
                    onSuccess = {
                        zkDocuments += it.toZkDocument()
                    },
                    onFailure = {
                        if (zkMetadata.zkInfo.zkRequired) {
                            throw it
                        } else {
                            plainDocuments += param.credential.discloseRequestedClaims(param.claims, request)
                                .getOrThrow()
                        }
                    }
                )
            } else {
                plainDocuments += param.credential.discloseRequestedClaims(param.claims, request).getOrThrow()
            }
        }

        return CreatePresentationResult.DeviceResponse(
            deviceResponse = DeviceResponse(
                parsedVersion = Version(1, 0),
                documents = plainDocuments.toTypedArray(),
                zkDocuments = zkDocuments.toTypedArray(),
                status = 0U,
            ),
        )
```

vck/src/commonTest/kotlin/at/asitplus/wallet/lib/agent/VerifiablePresentationFactorySdJwtTest.kt
```diff
@@ -63,7 +65,10 @@ val VerifiablePresentationFactorySdJwtTest by matrixSuite {
             ).getOrThrow()
 
             object {
-                val verifiablePresentationFactory = VerifiablePresentationFactory(holderKeyMaterial)
+                val verifiablePresentationFactory = VerifiablePresentationFactory(
+                    keyMaterial = holderKeyMaterial,
+                    mdocZkEngine = IsoMdocZkEngine(IsoMdocZkBackendRegistry())
```
nodh: Why do we need this in a test for SD-JWT?

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/zk/iso/IsoMdocZkEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+            ): Pair<IsoMdocZkBackend, List<ZkSystem>>? =
+                supportedBackends.entries.firstOrNull()?.toPair()
+
+            override fun selectForLoad(candidates: List<IsoMdocZkBackend>): IsoMdocZkBackend? =
+                candidates.firstOrNull()
+        }
+    }
+
+    companion object {
+        val Default = IsoMdocZkEngine()
```
nodh: Not needed: Just invoke the constructor without arguments

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/zk/iso/IsoMdocZkEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+import at.asitplus.catching
+import at.asitplus.iso.SessionTranscript
+import at.asitplus.iso.ZkDocument
+import at.asitplus.iso.ZkSystem
+import at.asitplus.wallet.lib.agent.IsoPresentationParameters
+import at.asitplus.wallet.lib.agent.PresentationException
+import at.asitplus.wallet.lib.agent.PresentationRequestParameters
+import at.asitplus.wallet.lib.agent.ZkMetadata
+
+class IsoMdocZkEngine(
```
nodh: Please add `@JvmOverloads` to the constructor

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/zk/iso/IsoMdocZkEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+    private val registry: IsoMdocZkBackendRegistry = IsoMdocZkBackendRegistry.Default,
+    private val selectionStrategy: SelectionStrategy = SelectionStrategy.Default
+) {
+    suspend fun generate(
+        request: PresentationRequestParameters,
+        isoParameters: IsoPresentationParameters
+    ): KmmResult<IsoMdocZkProof> = catching {
+        val zkMetadata = isoParameters.zkMetadata as? ZkMetadata.IsoMdocZk
+            ?: throw IllegalArgumentException("ZK metadata incompatible with Iso mDoc!")
```
nodh: no need for `!` in exception messages :-)

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/zk/iso/IsoMdocZkEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    fun load(
+        zkSystem: ZkSystem,
+        zkDocument: ZkDocument,
+        sessionTranscript: SessionTranscript
+    ): KmmResult<IsoMdocZkProof> = catching {
+        val zkSystemId = zkDocument.zkDocumentDataBytes.value.zkSystemId
+        if (zkSystemId != zkSystem.zkSystemId)
+            throw PresentationException("ZKSystemId mismatch")
```
nodh: please add the non-matching values to the exception message

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/zk/iso/IsoMdocZkEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+            ?: throw PresentationException("No backend found for ZK system: $zkSystemId")
+
+        backend.load(zkDocument, sessionTranscript, zkSystem)
+    }
+
+    private fun findSupportedBackends(
+        zkSystems: List<ZkSystem>,
+        backends: Set<IsoMdocZkBackend>
+    ): Map<IsoMdocZkBackend, List<ZkSystem>> =
+        backends
```
nodh: nitpick: move `backends` one line up to make the code more compact

nodh (changes requested):
Few small comments

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/PresentationDataClasses.kt
opsiv: We will need the SessionTranscript for zero-knowledge proof generation (and later for verification). 
Longfellow's ISO mDoc circuit uses the session transcript as a mandatory public parameter.
The DeviceSignature does not allow re-calculating the SessionTranscript


Are there other means to get the SessionTranscript?

vck/src/commonTest/kotlin/at/asitplus/wallet/lib/agent/ValidatorVpTest.kt
opsiv: No, this is a remnant from an earlier version without default values. I'll revert it. (same for SD-JWT Test)

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/PresentationDataClasses.kt
nodh: But this new parameter does not get set in `IsoMdocDcapiResponseBuilder`, and it never gets executed at all? Plus it looks weird in `PresentationFactory` now that `sessionTranscriptCallback` is used for both the `PresentationRequestParameters` and the `calcDeviceSignature`. Maybe it helps to add `sessionTranscriptCallback` to `IsoDeviceSignatureInput`?

opsiv: Yes nice catch! I overlooked setting it in `IsoMdocDcapiResponseBuilder`. 
The fact that the `calcIsoDeviceSignaturePlain` callback parameter in `IsoMdocDcapiResponseBuilder` or `PresentationFactory` incorporate a `SessionTranscript` could lead to undetected inconsistencies. This (combined with the fact that it isn't being used yet) makes this change feel uncomfortable.
In the most recent commit i made sure that the callback to calculate the transcript is being used for `calcIsoDeviceSignaturePlain` too, but architecturally it still feels a bit brittle.

Maybe I don't understand your suggestion to move it to `IsoDeviceSignatureInput`. I think the protocol (OID4VP/DC-API/Proximity) dependency on how `SessionTranscript` is calculated does complicate things too.
So does the fact that the callbacks for both `calcIsoDeviceSignaturePlain` and `calcIsoSessionTranscript` do not need to be overriden for non-mDoc flows, because they are never invoked in that case.

opsiv: On the other hand SessionTranscript does feel like it should belong into `PresentationRequestParameters`, just because it is a parameter for excatly one request.

nodh: As discussed offline we'll refactor that in a new PR

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/agent/VerifiablePresentationFactory.kt
nodh: your code is even better!

nodh (approved):

PR #642, opened 2026-08-12
OAuth: Rework token validation

nodh:
Reworks the validation of OAuth tokens, to prepare for DPoP-combined mode https://www.ietf.org/archive/id/draft-ietf-oauth-attestation-based-client-auth-10.html#name-dpop-combined-mode

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialIssuer.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialIssuer.kt
@@ -30,6 +30,7 @@ import at.asitplus.wallet.lib.jws.SignJwt
 import at.asitplus.wallet.lib.jws.SignJwtFun
 import at.asitplus.wallet.lib.jws.VerifyJwsObject
 import at.asitplus.wallet.lib.oauth2.RequestInfo
+import at.asitplus.wallet.lib.oauth2.ValidatedAccessToken
 import at.asitplus.wallet.lib.oidvci.OAuth2Exception.*
 import io.github.aakira.napier.Napier
 import kotlinx.serialization.json.Json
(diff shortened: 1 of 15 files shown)


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/CredentialIssuer.kt
```diff
@@ -247,42 +250,37 @@ class CredentialIssuer @JvmOverloads constructor(
...
             if (credentialIdentifier == null)
                 throw InvalidCredentialRequest("credential_identifier expected to be set")
             if (credentialConfigurationId != null)
                 throw InvalidCredentialRequest("credential_configuration_id must not be set when credential_identifier is set")
-            if (!it.validCredentialIdentifiers.contains(credentialIdentifier))
-                throw InvalidToken("credential_identifier $credentialIdentifier expected to be in $it")
-        } else if (it.scope != null) {
+            if (!token.validCredentialIdentifiers.contains(credentialIdentifier))
+                throw InvalidToken("credential_identifier $credentialIdentifier expected to be in $token")
+        } else if (token.scope != null) {
```
JesusMcCloud, line 273: i am asking because i really don't know (and yes, not new logic, i know): why is this mutually exclusive?

nodh: Because it's defined in [OID4VCI](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-request) that way:

> In the credential request ... `credential_identifier` REQUIRED when authorization details was returned in the token, MUST NOT be used otherwise ... `credential_configuration_id` REQUIRED when no authorization details were returned in the token (that's the `scope` case)

JesusMcCloud (approved):

PR #646, opened 2026-08-14
OAuth: Bind client authentication to internal state

nodh:
Bind client authentication to the internal state, so that we can't confuse authenticated clients across requests.

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/SimpleAuthorizationService.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/SimpleAuthorizationService.kt
@@ -317,15 +317,16 @@ class SimpleAuthorizationService @JvmOverloads constructor(
     ) = catching {
         val actualRequest = request.extractPushedRequestParams()
         Napier.i("par called with $actualRequest")
-        val authenticatedClient = clientAuthenticationService
+        val client = clientAuthenticationService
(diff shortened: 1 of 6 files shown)


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oauth2/SimpleAuthorizationService.kt
```diff
@@ -706,3 +738,8 @@ data class ResponseWithDpopNonce<T>(
     val response: T,
     val dpopNonce: String?,
 )
+
+data class PushedAuthorizationRequest(
```
n0900, line 742: Can this be internal? / misleading name?
PAR just allows to push `AuthenticationRequestParameters` or a `RequestObject` (==`JwsSigned<AuthenticationRequestParameters>`) to an `/par` endpoint but doesnt define a new format itself right?

nodh: Should be internal, but then implementers of `MapStore` (constructor parameter) can't use this to serialize it or do anything else.

n0900: We should still probably rename/document that this is not a true PAR class but only an internal model.

JesusMcCloud (approved):
lgtm

PR #649, opened 2026-08-17
OAuth: Implement DPoP combined mode for client authentication

nodh:
Final PR in the large stack: Actually implement DPoP combined mode defined in [OAuth 2.0 Attestation-Based Client Authentication](https://www.ietf.org/archive/id/draft-ietf-oauth-attestation-based-client-auth-10.html#name-dpop-combined-mode) referencing [OAuth 2.0 Demonstrating Proof of Possession (DPoP)](https://datatracker.ietf.org/doc/html/rfc9449)


JesusMcCloud (approved):
Looking good, but I have yet to reach the top of the stack

PR #651, opened 2026-08-17
OpenID4VP: Correctly use ephemeral encryption keys

nodh:
Asserts conformance to [HAIP 1.0](https://openid.net/specs/openid4vc-high-assurance-interoperability-profile-1_0-final.html#section-5-2.6):
> Verifiers MUST supply ephemeral encryption public keys specific to each Authorization Request passed via client metadata as specified in Section 8.3 of [[OIDF.OID4VP](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-encrypted-responses)].

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/DcApiVerifier.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/DcApiVerifier.kt
@@ -37,6 +37,7 @@ import at.asitplus.signum.supreme.sign.Signer
 import at.asitplus.wallet.lib.DefaultNonceService
 import at.asitplus.wallet.lib.MdocDeviceSignatureVerifier
 import at.asitplus.wallet.lib.NonceService
+import at.asitplus.wallet.lib.agent.EphemeralEncryptionKeyService
 import at.asitplus.wallet.lib.agent.EphemeralKeyWithoutCert
 import at.asitplus.wallet.lib.agent.KeyMaterial
 import at.asitplus.wallet.lib.agent.NonceChallengeVerifier
@@ -47,8 +48,8 @@ import at.asitplus.wallet.lib.cbor.VerifyCoseSignatureWithKey
 import at.asitplus.wallet.lib.cbor.VerifyCoseSignatureWithKeyFun
 import at.asitplus.wallet.lib.data.CredentialPresentationRequest.DCQLRequest
 import at.asitplus.wallet.lib.data.CredentialPresentationRequest.IsoDeviceRetrieval
-import at.asitplus.wallet.lib.jws.DecryptJwe
 import at.asitplus.wallet.lib.jws.DecryptJweFun
+import at.asitplus.wallet.lib.jws.DecryptJweWithEphemeralKey
 import at.asitplus.wallet.lib.jws.SignJwt
 import at.asitplus.wallet.lib.jws.SignJwtFun
 import at.asitplus.wallet.lib.jws.VerifyJwsObject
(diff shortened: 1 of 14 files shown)


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/DcApiVerifier.kt
```diff
@@ -79,10 +80,18 @@ class DcApiVerifier @JvmOverloads constructor(
...
-    private val decryptionKeyMaterial: KeyMaterial = EphemeralKeyWithoutCert(),
-    /** Decrypts encrypted responses from holders. */
-    private val decryptJwe: DecryptJweFun = DecryptJwe(decryptionKeyMaterial),
+    /**
+     * Long-lived key advertised in [metadata] so that holders can encrypt responses, but **only** for client
+     * identifier schemes that do not convey client metadata in the request, i.e. where this key is distributed
+     * out-of-band. This is not conformant to OpenID4VC HAIP, so leave it `null` to have every request carry its own
+     * ephemeral encryption key, see [ephemeralEncryptionKeyService].
+     */
+    private val decryptionKeyMaterial: KeyMaterial? = null,
```
JesusMcCloud, line 89: Ah, yes…

JesusMcCloud (approved):

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/jws/JwsService.kt
```diff
@@ -455,6 +456,26 @@ class DecryptJwe(
 
 }
 
+/**
+ * Decrypts JWE payloads with the ephemeral key referenced by the JWE's `kid` header, as per
+ * [OpenID4VP 1.0, 8.3](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-encrypted-responses).
+ */
+class DecryptJweWithEphemeralKey(
```
JesusMcCloud, line 463: This key is never consumed. only `getKey()` is invoked.  `getKey()` function should have an opt-in requirement

JesusMcCloud (changes requested):

tlenz (approved):
From an architectural perspective, the implementation should meet the requirements of OpenID4VP and HAIP. Multi-node deployments are supported through the MapStore concept, which enables the required state to be shared across nodes.

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/DcApiVerifier.kt
nodh: See https://github.com/a-sit-plus/vck/issues/654 as discussed offline

vck/src/commonMain/kotlin/at/asitplus/wallet/lib/jws/JwsService.kt
nodh: I've documented it. Get without consume has a safety net due to the expiration of entries in `DefaultMapStore` (10 minutes), so we do not accumulate keys unbounded.

JesusMcCloud: still, why don't we cosume it here?

nodh: Reworked that part so that the key is immediately consumed. Decided against trying to decrypt first and then consume, since the keys are shortlived anyway.

JesusMcCloud (approved):
lgtm. maybe just plainly document in the code the (theoritical) attack vector. this underscores that the design is deliberate

PR #653, opened 2026-08-19
OpenID4VP: Support fetching encrypted authn requests

nodh:
Implements a missing feature: In OpenID4VP verifiers can support `POST` requests to deliver the Authn Request. In those requests the wallet may send a key that the verifier should use to encrypt the authn request for. See [OpenID4VP 5.10](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-request-uri-method-post).

### DIFF
--- a/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/SerializerSketch.kt
+++ b/vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/SerializerSketch.kt
@@ -1,9 +1,14 @@
 package at.asitplus.wallet.lib.oidvci
 
+import at.asitplus.catchingUnwrapped
 import at.asitplus.signum.indispensable.josef.io.joseCompliantSerializer
 import io.ktor.http.*
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerializationException
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.CompositeDecoder
+import kotlinx.serialization.serializer
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonArray
 import kotlinx.serialization.json.JsonObject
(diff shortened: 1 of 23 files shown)


vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/oidvci/SerializerSketch.kt
```diff
@@ -15,14 +20,28 @@ import kotlinx.serialization.json.encodeToJsonElement
...
+        // ponytail: `v[0]` throwing on an empty value is load-bearing: `ResponseParser` relies on it to fall back
+        // from parsing an input as URL to parsing it as POST body
+        k to when {
+            // members declared as strings keep their content verbatim, even if that content happens to be JSON
+            // itself, e.g. `wallet_metadata` of `at.asitplus.openid.RequestObjectParameters`
+            descriptor.isStringElement(k) -> JsonPrimitive(v)
+            v[0] == '{' -> json.decodeFromString<JsonObject>(v)
+            v[0] == '[' -> json.decodeFromString<JsonArray>(v)
             else -> JsonUnquotedLiteral(v)  //no quoted → can be any type for deserializing. requires lenient parsing
         }
```
JesusMcCloud: i don't like this being reified by default. rather: make one taking a serializer/descriptor and then add a reified one that delegates; this way we can still manually specify serializers cleanly.

Also: what is that json instnace? I'd rather pass it explicitly as parameter or (probably better) use `context(json: Json)`

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/OpenId4VpRequestFactory.kt
```diff
@@ -156,6 +164,53 @@ internal class OpenId4VpRequestFactory(
         ).getOrThrow()
     }
 
+    /**
+     * Creates the request object to serve at the `request_uri` endpoint: a signed request object, encrypted to the
+     * wallet's key if it passed one in its `wallet_metadata`, as per
+     * [OpenID4VP 1.0, 5.10](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-request-uri-method-post).
+     */
+    suspend fun createRequestObject(
```
JesusMcCloud, line 172: this method and its siblings are due for a name overhaul. not saying it must be done here, but we have:

* createPlainAuthnRequest
* createSignedRequestObject
* createRequestObject

that are nested in that order. not very clear, but also not urgent, because it is only on an internal class

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/OpenId4VpRequestFactory.kt
```diff
@@ -156,6 +164,53 @@ internal class OpenId4VpRequestFactory(
...
+                recipientKey = encryptionTarget.first,
+            ).getOrThrow().serialize()
+        }
+    }
+
+    /**
+     * The wallet's encryption key and the content encryption algorithm to use, or `null` if the wallet did not ask for
+     * an encrypted request object, or asked for algorithms we don't support.
+     */
+    private fun OAuth2AuthorizationServerMetadata.encryptionTarget(): Pair<JsonWebKey, JweEncryption>? {
```
JesusMcCloud: since this only references immutable properties, dubsequent calls will be idempotent. hence i'd refactor it into an extn property, not a function

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/RequestParser.kt
```diff
@@ -84,33 +103,94 @@ class RequestParser(
...
-    } ?: parameters.requestUri
-        ?.let { remoteResourceRetriever.invoke(parameters.resourceRetrieverInput(it)) }
-        ?.let {
-            it.parseAsJwsRequest(parent)
+    } ?: parameters.requestUri?.let { uri ->
+        val method = parameters.requestUriMethod?.toHttpMethod() ?: HttpMethod.Get
+        // only the POST request to the request URI endpoint has a channel for these, see OpenID4VP 1.0, 5.10, and it
+        // is built once per request, since it carries the key the verifier shall encrypt this very request to
+        val requestObjectParameters = if (method == HttpMethod.Post) buildRequestObjectParameters.invoke() else null
+        remoteResourceRetriever.invoke(parameters.resourceRetrieverInput(uri, method, requestObjectParameters))?.let {
```
JesusMcCloud: invoke literally spelled out intentionally? i don't have a local copy at hand, at the moment. feel free to ignore if my comment makes no sense

JesusMcCloud (changes requested):

vck-openid/src/commonMain/kotlin/at/asitplus/wallet/lib/openid/OpenId4VpRequestFactory.kt
nodh: Well, if you're optionally encrypting a signed request wrapping a plain request, so there has to be some form of nesting. But I've refactored out the public interface in `OpenId4VpVerifier` with these names.

JesusMcCloud: i did not mind the nesting, but naming inconsistincy and potential confusion

JesusMcCloud: still: much better now!

vck-openid/src/commonTest/kotlin/at/asitplus/wallet/lib/oidvci/SerializationTest.kt
```diff
@@ -88,7 +88,7 @@ val SerializationTest by matrixSuite {
             this shouldContain "dcql_query=" + "{\"credentials".encodeURLParameter()
         }
 
-        intermediateMap.decode<AuthenticationRequestParameters>() shouldBe params
+        intermediateMap.decode(AuthenticationRequestParameters.serializer()) shouldBe params
```
JesusMcCloud, line 91: should work reified with the extension, but i don't mind either way

JesusMcCloud (approved):
🚢 🇮🇹

PR #657, opened 2026-08-20
OAuth: More hardening for the AS

nodh:
Not sure if this PR is too large, on the other hand five PR for five single commits also seems too much.


JesusMcCloud (approved):
lgtm

PR #658, opened 2026-08-21
OID4VP: Harden request parsing

nodh:
Error in Valera can be triggered by scanning QR Code in profile "Generic EUDI Wallet" from https://eudiw-playground.eideasy.com/qr ... but that contains several fields that should not be there, but tripped Valera into a ugly looking class cast exception. This commit here shows a better error message.


JesusMcCloud (approved):
lgtm

n0900 (approved):

PR #660, opened 2026-08-26
ISO mDoc ZK: align backend APIs with ZkRequest and ZkSystemSpec

opsiv:
- Replace the preparation-only ZkInfo and ZkSystem abstractions with the values used internally for proof generation.
- Make ZkRequest validate required systems and unique system IDs
- Rename ZkSystemSpec.zkSystemId to id
- Convert DCQL ZK metadata into ZkRequest
- Widen DCQL ZK numeric parameters from Int to Long (to stay consistent with other implementations like Multipaz)
- Update backend, engine, and proof APIs to use ZkSystemSpec
- Pass holder key material into backend proof generation (was incorrectly missing before)
- Wire the revised ZK metadata through holder presentation creation
- Update architecture documentation and the 8.0 changelog


nodh:
@codex review

openid-data-classes/src/commonMain/kotlin/at/asitplus/iso/ZkSystemSpecSerializer.kt
```diff
@@ -51,7 +51,7 @@ object ZkSystemSpecSerializer : KSerializer<ZkSystemSpec> {
         }
 
         return ZkSystemSpec(
-            zkSystemId = zkSystemId ?: error("Missing zkSystemId"),
+            id = zkSystemId ?: error("Missing zkSystemId"),
```
nodh: Maybe worth to rename `zkSystemId` here to `id` too (also in the message)?

nodh (approved):

nodh (approved):

PR #666, opened 2026-08-31
Deprecate presentation exchange

gp-iaik:
Follow up to https://github.com/a-sit-plus/vck/pull/608


nodh (changes requested):
I'd prefer to keep the additional `@Deprecated` annotations from this PR, but remove the `require` statements, and remove everything deprecated with 9.0 (then also the `require` statements are not needed).

